PR #2, opened 2023-04-18
ACTIN-4 Extract known genes from CKB and Hartwig

pauldwolfe:
First steps, did most of the plumbing in datamodel and algo.

In CKB, had a first try at consolidating all the gene info
present in the current EventExtractorOutput. Haven't looked
at the filtering yet.


algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -307,4 +310,9 @@ private static Set<KnownFusion> convertToKnownFusions(@Nullable FusionPair fusio
 
         return FusionConsolidation.consolidate(fusions);
     }
+
+    @NotNull
+    private static Set<KnownGene> convertToKnownGene(@NotNull String gene) {
+        return Set.of(ImmutableKnownGene.builder().gene(gene).geneRole(GeneRole.ONCO).build());
```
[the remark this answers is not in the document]
kduyvesteyn: Sharp :) I meant to say "Should be set to `UNKNOWN`" !!

algo/src/main/java/com/hartwig/serve/ServeAlgo.java
```diff
@@ -173,13 +184,35 @@ private ExtractionResult extractHartwigCohortKnowledge(@NotNull String hartwigCo
...
+        HartwigHotspotExtractor extractor =
+                new HartwigHotspotExtractor(Knowledgebase.HARTWIG_CURATED, refGenomeResource.proteinResolver(), addExplicitHotspots);
         LOGGER.info("Running Hartwig Curated knowledge extraction");
         return extractor.extract(entries);
     }
+
+    @NotNull
+    private ExtractionResult extractHartwigDriverGeneKnowledge(@NotNull String driverGeneFileTsv) throws IOException {
+        LOGGER.info("Reading Hartwig Driver Genes TSV from '{}'", driverGeneFileTsv);
+        List<HartwigGeneEntry> entries = HartwigGeneFileReader.read(driverGeneFileTsv);
```
pauldwolfe, line 200: we could also read directly from the `DriverGeneFile` here and just pluck out the gene. Perhaps that's more robust, but felt inconsistent. What do you think @kduyvesteyn ?

kduyvesteyn: I prefer your solution. There is a somewhat sneaky dependency that both input files have a column called 'gene' but I think that's a fairly robust assumption anyways (also considering both files are managed by HMF).

algo/src/main/java/com/hartwig/serve/ServeConfig.java
```diff
@@ -149,6 +149,12 @@ static Options createOptions() {
     @NotNull
     String hartwigCuratedTsv();
 
+    boolean useHartwigDriverGenes();
+
+    boolean useHartwigCuratedGenes();
+
+    String hartwigCuratedGeneTsv();
```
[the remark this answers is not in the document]
kduyvesteyn: See mail :)

algo/src/test/java/com/hartwig/serve/ServeAlgoTest.java
```diff
@@ -26,6 +26,7 @@ public class ServeAlgoTest {
 
     private static final String DOCM_TSV = Resources.getResource("docm/example.tsv").getPath();
     private static final String HARTWIG_CURATED_TSV = Resources.getResource("hartwig/example.tsv").getPath();
+    private static final String HARTWIG_CURATED_GENES_TSV = Resources.getResource("hartwig/example.tsv").getPath();
```
kduyvesteyn, line 29: Suggest to structure the test resources the same as production resources (in crunch-resources-private)

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbVariantAnnotatorTest.java
kduyvesteyn, line 1: A little embarrassing this didn't exist yet. Must have certainly rushed it :(

kduyvesteyn (commented):
- I have made a few small changes (see commits)
 - The implementation of hartwig genes looks all good to me. I would rename the previous "hartwig" to "hartwig hotspots" all the way though (also rename the input params, Knowledgebase enum values etc. It must be clear all the way that we provide inputs to both hotspots and genes, and that both are independent. 
 - CKB handling looks correct to me now.

PR #4, opened 2023-04-24
ACTIN-82 Include trailing empty strings from known fusions

pauldwolfe:
Looks like some recent changes broke the known fusions file, the previous high promiscuous impact field container NA
and now it is empty string. This breaks the parsing which did a split with now limit option


kduyvesteyn (approved):
Changes look good. Have expanded on the tests themselves (fixed a TODO).

PR #9, opened 2023-12-04
ACTIN-68: Develop SERVE CKB trials importer

cbruel:
https://hartwigmedical.atlassian.net/browse/ACTIN-68

Enjoy your peppernoten :)


pauldwolfe (approved):

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractorFactory.java
```diff
@@ -14,10 +15,20 @@ private CkbExtractorFactory() {
...
 
     @NotNull
-    public static CkbExtractor buildCkbExtractor(@NotNull EventClassifierConfig config, @NotNull RefGenomeResource refGenomeResource,
+    public static CkbExtractor createEvidenceExtractor(@NotNull EventClassifierConfig config, @NotNull RefGenomeResource refGenomeResource,
             @NotNull TreatmentApproachCurator treatmentApproachCurator) {
         // We want to capture all events from CKB, so ignore driver inconsistencies
-        return new CkbExtractor(EventExtractorFactory.create(config, refGenomeResource, DriverInconsistencyMode.IGNORE),
-                treatmentApproachCurator);
+        return new CkbExtractor(Knowledgebase.CKB_EVIDENCE,
+                EventExtractorFactory.create(config, refGenomeResource, DriverInconsistencyMode.IGNORE),
```
kduyvesteyn, line 22: I suggest to create the event extractor in a separate function that is reused between both evidence and trial extractor creation (along with the comment on why we set the driver inconsistency mode to `IGNORE`). This way you are guaranteed that event extraction is always identical between trials and evidence

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableFunctions.java
```diff
@@ -0,0 +1,102 @@
...
+                default: {
+                    // CKB uses 10000005 for configuring "Not a cancer". We can ignore these.
+                    if (!id.equals(CancerTypeConstants.JAX_NOT_CANCER)) {
+                        LOGGER.warn("Unexpected DOID string annotated by CKB: '{}'", source + ":" + id);
+                    }
+                    return null;
+                }
+            }
+        } else {
+            LOGGER.warn("Unexpected source '{}'", source);
```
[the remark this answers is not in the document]
kduyvesteyn: SERVE output can be used in case this situation occurs. The WARNs in SERVE are fairly similar to the warns in ACTIN-clinical, ideally they should be captured in a single source and handed differently (not relying on a human reading the command line output). I'd say for now lets leave these WARNs, as long as we have no intention to go full CICD with SERVE

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -64,14 +67,20 @@ public class CkbExtractor {
     private static final Logger LOGGER = LogManager.getLogger(CkbExtractor.class);
     private static final String DELIMITER = ",";
```
kduyvesteyn, line 68: Not related to your changes but would rename this to `VARIANT_DELIMITER`

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
kduyvesteyn, line 1: The progress tracker on line 90 currently has hardcoded `CKB`, might be nice to replace with `source` so that in the logs we can differentiate between evidence and trials?

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -175,23 +186,37 @@ private static ExtractionResult toExtractionResult(@NotNull String variant, @Not
...
+                    .addEventInterpretations(interpretation)
+                    .actionableHotspots(actionableHotspots)
+                    .actionableCodons(actionableCodons)
+                    .actionableExons(actionableExons)
+                    .actionableGenes(actionableGenes)
+                    .actionableFusions(actionableFusions)
+                    .actionableCharacteristics(actionableCharacteristics)
+                    .actionableHLA(actionableHLA)
+                    .build();
+        }
```
kduyvesteyn, line 219: There is duplication here!

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -175,23 +186,37 @@ private static ExtractionResult toExtractionResult(@NotNull String variant, @Not
...
+                    .actionableCodons(actionableCodons)
+                    .actionableExons(actionableExons)
+                    .actionableGenes(actionableGenes)
+                    .actionableFusions(actionableFusions)
+                    .actionableCharacteristics(actionableCharacteristics)
+                    .actionableHLA(actionableHLA)
+                    .build();
+        } else {
+            return ImmutableExtractionResult.builder()
+                    .refGenomeVersion(Knowledgebase.CKB_EVIDENCE.refGenomeVersion())
```
kduyvesteyn, line 209: This should use the `refGenomeVersion` of the source property.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -229,7 +254,7 @@ private static Set<KnownHotspot> convertToKnownHotspots(@Nullable List<VariantHo
                         .from(hotspot)
                         .geneRole(GeneRole.UNKNOWN)
                         .proteinEffect(ProteinEffect.UNKNOWN)
-                        .addSources(Knowledgebase.CKB)
+                        .addSources(Knowledgebase.CKB_EVIDENCE)
```
kduyvesteyn, line 257: Also here (and in subsequent usages) you should use the `source` property

algo/src/main/java/com/hartwig/serve/sources/ckb/treatmentapproach/TreatmentApproachCurator.java
```diff
@@ -31,7 +30,7 @@ public void reportUnusedCuratedEntries() {
             if (!usedCurations.contains(entrySet)) {
                 unusedCuratedEntryCount++;
                 // TODO Enable warning once treatment approaches are properly curated.
-//                LOGGER.warn(" Curation entry '{}' hasn't been used for treatment approach curation", entrySet);
```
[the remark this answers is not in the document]
kduyvesteyn: This is code written by Lieke, not sure what the status is (but was unchanged in this ticket)

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -0,0 +1,109 @@
...
+        POTENTIALLY_OPEN_RECRUITMENT_TYPES.add("RECRUITING");
+        POTENTIALLY_OPEN_RECRUITMENT_TYPES.add("ACTIVE_NOT_RECRUITING");
+
+        COUNTRIES_TO_INCLUDE.add("Netherlands");
+        COUNTRIES_TO_INCLUDE.add("Belgium");
+        COUNTRIES_TO_INCLUDE.add("Germany");
+
+        VARIANT_REQUIREMENT_TYPES_TO_INCLUDE.add("partial - required");
+        VARIANT_REQUIREMENT_TYPES_TO_INCLUDE.add("required");
+        VARIANT_REQUIREMENT_TYPES_TO_INCLUDE.add("partial");
```
kduyvesteyn, line 40: I'd tend to think it would be better to make this case insensitive?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableFunctionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+
+import com.google.common.collect.Sets;
+import com.hartwig.serve.cancertype.CancerTypeConstants;
+
+import org.junit.Test;
+
+public class ActionableFunctionsTest {
```
kduyvesteyn, line 12: Missing line break

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableFunctionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+        assertNull(ActionableFunctions.extractCancerTypeDetails(CkbTestFactory.createIndication("test", "JAX:10000004")));
+
+        assertEquals(Sets.newHashSet(CancerTypeConstants.REFRACTORY_HEMATOLOGIC_TYPE,
+                        CancerTypeConstants.BONE_MARROW_TYPE,
+                        CancerTypeConstants.LEUKEMIA_TYPE),
+                ActionableFunctions.extractCancerTypeDetails(CkbTestFactory.createIndication("test", "JAX:10000003"))
+                        .blacklistedCancerTypes());
+        assertEquals(Sets.newHashSet(),
+                ActionableFunctions.extractCancerTypeDetails(CkbTestFactory.createIndication("test", "JAX:10000009"))
+                        .blacklistedCancerTypes());
```
kduyvesteyn, line 40: I would use something like `assertTrue(isEmpty)`

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableFunctionsTest.java
```diff
@@ -0,0 +1,49 @@
...
+
+    }
+
+    private String createCancerTypeDetails(String termId) {
+        return ActionableFunctions.extractCancerTypeDetails(CkbTestFactory.createIndication("test", termId))
+                .applicableCancerType()
+                .doid();
```
kduyvesteyn, line 47: This function can be static, and is missing nullable annotations.

Also the name can be clearer I think. You are extracting the doid of the applicable cancer type.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+                "KRAS deletion",
+                "sensitive",
+                "Emerging",
+                "AB",
+                "AB",
+                "A",
+                "DOID:162",
+                location,
+                "Recruiting",
+                requirementType);
```
kduyvesteyn, line 41: I would make a simpler utility function. For trials you don't care about evidence so that can be omitted from this utility function.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+import org.apache.logging.log4j.util.Strings;
+import org.junit.Test;
+
+public class ActionableTrialFactoryTest {
+
+    @Test
+    public void canCreateActionableTrials() {
+        Location location = ImmutableLocation.builder().nctId("").city("").country("United States").status("Recruiting").build();
+        VariantRequirementDetail requirementType =
+                ImmutableVariantRequirementDetail.builder().profileId(0).requirementType("required").build();
```
kduyvesteyn, line 29: I would suggest to add utility functions for this

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+    @Test
+    public void canDetermineIfHasCountryToIncludeWithPotentiallyOpenRecruitmentType() {
+        ClinicalTrial notOpenDutchTrial =
+                CkbTestFactory.createTrialWithCountryAndRecruitmentType(Lists.newArrayList(ImmutableLocation.builder()
+                        .nctId("")
+                        .city("")
+                        .country("Netherlands")
+                        .status("Not yet recruiting")
+                        .build()), "Recruiting");
+        ClinicalTrial OpenDutchTrial =
```
kduyvesteyn, line 130: Should be `openDutchTrial`

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
kduyvesteyn, line 1: Would be good to discuss the setup of this test in more detail face-to-face

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
kduyvesteyn, line 1: The test should be more clear that you only test evidence extraction.

I think it would be good to also test trial extraction, eg: make it clear no known events are generated from trials.

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbTestFactory.java
```diff
@@ -78,6 +85,55 @@ public static CkbEntry createEntryWithGeneAndVariant(@NotNull String geneSymbol,
...
+                        recruitmentType,
+                        indicationName,
+                        termId,
+                        List.of(requirementType),
+                        List.of(location))))
+                .build();
+    }
+
+    @NotNull
+    public static ClinicalTrial createTrialWithCountryAndRecruitmentType(@NotNull List<Location> locations, String recruitmentType) {
```
kduyvesteyn, line 110: Missing nullable annotation

datamodel/src/main/java/com/hartwig/serve/datamodel/Knowledgebase.java
```diff
@@ -12,6 +12,8 @@
 
 public enum Knowledgebase {
     CKB(RefGenome.V38, "CKB", EvidenceLevel.C, EvidenceLevel.B),
```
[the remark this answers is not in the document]
kduyvesteyn: Good one, done.

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/BackwardsCompatibilityUtil.java
```diff
@@ -0,0 +1,186 @@
...
+import com.hartwig.serve.datamodel.range.ImmutableKnownCodon;
+import com.hartwig.serve.datamodel.range.ImmutableKnownExon;
+import com.hartwig.serve.datamodel.range.KnownCodon;
+import com.hartwig.serve.datamodel.range.KnownCodonComparator;
+import com.hartwig.serve.datamodel.range.KnownExon;
+import com.hartwig.serve.datamodel.range.KnownExonComparator;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class BackwardsCompatibilityUtil {
```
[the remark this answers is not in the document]
kduyvesteyn: You are absolutely right, and not your fault since I also didn't think of it. 

An advantage of my current solution would be that OncoAct can simply bump the datamodel version without making any code changes, but if there are allowed to do that, then they may as well make the code change itself. Have asked Lieke for input. 

Agree the only way for true backwards compatibility would be to write the data as CKB. We could duplicate the data completely or add a flag (OncoAct uses a SERVE database from their own process anyways, and could set the flag).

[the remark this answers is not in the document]
pauldwolfe: The flag is a good idea. That's actually a pretty normal pattern (feature toggle) for this sort of thing, although I would have the default as CKB such that you "opt-in" to the new feature.

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
cbruel: added the testing of trial extraction

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
cbruel: I've removed the evidence stuff from the ActionableTrialFactoryTest

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
cbruel: Done!

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
cbruel: Learned from you how I could solve this :)

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
cbruel: Changed this to showing CKB_EVIDENCE and CKB_TRIAL

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -0,0 +1,109 @@
...
+                countries.add(location.country());
+            }
+        }
+        return countries;
+    }
+
+    @VisibleForTesting
+    static boolean hasVariantRequirementTypeToInclude(@NotNull List<VariantRequirementDetail> variantRequirementDetails, CkbEntry entry) {
+        for (VariantRequirementDetail variantRequirementDetail : variantRequirementDetails) {
+            if (entry.profileId() == variantRequirementDetail.profileId() && VARIANT_REQUIREMENT_TYPES_TO_INCLUDE.contains(
```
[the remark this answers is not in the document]
cbruel: as discussed, this is necessary (trial can have multiple molecularProfiles)

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -0,0 +1,109 @@
...
+    @Override
+    public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sourceEvent, @NotNull String gene) {
+        Set<ActionableEntry> actionableTrials = Sets.newHashSet();
+
+        for (ClinicalTrial trial : trialsToInclude(entry)) {
+            for (Indication indication : trial.indications()) {
+                CancerTypeExtraction cancerTypeExtraction = ActionableFunctions.extractCancerTypeDetails(indication);
+
+                if (cancerTypeExtraction != null) {
+                    Set<String> countries = countriesToInclude(trial);
```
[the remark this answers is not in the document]
cbruel: changed this

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractorFactory.java
cbruel: changed this

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableFunctions.java
```diff
@@ -0,0 +1,102 @@
...
+        String[] sourceCancerTypeDetails = extractSourceCancerTypeDetails(indication.termId());
+
+        if (sourceCancerTypeDetails == null) {
+            return null;
+        }
+
+        ImmutableCancerType.Builder applicableCancerTypeBuilder = ImmutableCancerType.builder().name(indication.name());
+        Set<CancerType> blacklistedCancerTypes = Sets.newHashSet();
+
+        assert sourceCancerTypeDetails.length == 2;
```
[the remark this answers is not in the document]
cbruel: Changed this to an illegalStateException, thanks for the suggestion!

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -0,0 +1,207 @@
...
+        List<Evidence> filtered = Lists.newArrayList();
+        for (Evidence evidence : evidences) {
+            if (hasUsableEvidenceType(evidence.evidenceType())) {
+                filtered.add(evidence);
+            }
+        }
+        return filtered;
+    }
+
+    @VisibleForTesting
```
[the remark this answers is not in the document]
kduyvesteyn: Didn't discuss extensively but this annotation and type of coding is quite omni-present in SERVE code base, feels odd to refactor a few cases. Also, this partial class is quite well readable so not a huge concern.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableFunctions.java
```diff
@@ -0,0 +1,105 @@
...
+        String[] sourceCancerTypeDetails = extractSourceCancerTypeDetails(indication.termId());
+
+        if (sourceCancerTypeDetails == null) {
+            return null;
+        }
+
+        ImmutableCancerType.Builder applicableCancerTypeBuilder = ImmutableCancerType.builder().name(indication.name());
+        Set<CancerType> blacklistedCancerTypes = Sets.newHashSet();
+
+        if (!(sourceCancerTypeDetails.length == 2)) {
```
kduyvesteyn, line 32: It would be a bit simpler to write `sourceCancerTypeDetails.length != 2`

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/BackwardsCompatibilityUtil.java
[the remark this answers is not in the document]
kduyvesteyn: After discussion with Lieke have decided to completely remove my attempt at creating backwards compatible. Its fine if this release brakes downstream components.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableFunctions.java
```diff
@@ -0,0 +1,102 @@
...
+import com.hartwig.serve.ckb.datamodel.indication.Indication;
+import com.hartwig.serve.datamodel.CancerType;
+import com.hartwig.serve.datamodel.ImmutableCancerType;
+
+import org.apache.logging.log4j.LogManager;
+import org.apache.logging.log4j.Logger;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+class ActionableFunctions {
```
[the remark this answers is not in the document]
kduyvesteyn: Also added a private constructor to make sure no one does try to instantiate an object of this class. This is no longer convention in ACTIN but it is used quite widely in SERVE.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+import com.hartwig.serve.sources.ckb.treatmentapproach.TreatmentApproachCurator;
+import com.hartwig.serve.sources.ckb.treatmentapproach.TreatmentApproachTestFactory;
+
+import org.apache.logging.log4j.util.Strings;
+import org.junit.Test;
+
+public class ActionableEvidenceFactoryTest {
+
+    @Test
+    public void canCreateActionableEntries() {
```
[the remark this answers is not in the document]
pauldwolfe: I still think you can go further. I see three tests in one here.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,147 @@
...
+import com.hartwig.serve.datamodel.EvidenceLevel;
+import com.hartwig.serve.datamodel.Knowledgebase;
+
+import org.apache.logging.log4j.util.Strings;
+import org.junit.Test;
+
+public class ActionableTrialFactoryTest {
+
+    @Test
+    public void canCreateActionableTrials() {
```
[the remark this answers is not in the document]
pauldwolfe: Is this still in progress? It also is still multiple tests in one unit test.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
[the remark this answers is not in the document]
cbruel: Hey Paul. I don't understand what you mean. Could you specify where you see three tests in one in ActionableEvidenceFactoryTest?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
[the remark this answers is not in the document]
cbruel: @pauldwolfe same here, could you specify what part of canCreateActionableTrial()? :)

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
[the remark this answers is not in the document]
pauldwolfe: Here too, github fail :)

pauldwolfe (approved):

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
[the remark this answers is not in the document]
cbruel: @pauldwolfe no problem! comments on outdated files are quite confusing indeed

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
kduyvesteyn, line 1: I suggest to make 2 separate functions to test evidence extraction and trial extraction. In addition, this would allow the removal of one glorified "createEntry" function in CkbTestFactory that creates an entry with both evidence and a trial (with a therapy name that is not really used elsewhere)

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+import com.hartwig.serve.datamodel.Knowledgebase;
+
+import org.apache.commons.compress.utils.Sets;
+import org.jetbrains.annotations.NotNull;
+import org.junit.Test;
+
+public class ActionableTrialFactoryTest {
+
+    @Test
+    public void canCreateActionableTrial() {
```
kduyvesteyn, line 28: I would rename this function to be more clear on what you are testing. Something like `canCreateActionableEntryForOpenTrialInAllowedCountry` or something similar.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        Location location = ImmutableLocation.builder().nctId("nctid").city("").country("Netherlands").status("Recruiting").build();
```
kduyvesteyn, line 29: You do create Locations in multiple places, always with an empty city and sometimes with an empty nctid. I recommend adding a `createLocation` function in CkbTestFactory to make this simpler

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        VariantRequirementDetail requirementType =
+                ImmutableVariantRequirementDetail.builder().profileId(0).requirementType("required").build();
+        CkbEntry entry = CkbTestFactory.createEntryWithClinicalTrial(location, "Recruiting", requirementType);
+        ActionableTrialFactory actionableTrialFactory = new ActionableTrialFactory();
+        Set<ActionableEntry> entrySet = actionableTrialFactory.create(entry, "KRAS", "gene");
```
kduyvesteyn, line 34: Would rename to `trials`

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        ClinicalTrial trialNotOpenInNetherlands = createTrial("Recruiting", "Netherlands", "Not yet recruiting");
+        ClinicalTrial trialPotentiallyOpenInUnitedStates = createTrial("Recruiting", "United States", "Recruiting");
+        ClinicalTrial trialWithdrawnInUnitedStates = createTrial("Recruiting", "United States", "Withdrawn");
+        ClinicalTrial trialPotentiallyOpenInBelgiumAndIndia =
+                createTrialWithMultipleLocations("Recruiting", "Belgium", "Recruiting", "India", "Recruiting");
+        ClinicalTrial trialPotentiallyOpenInBelgiumAndGermany =
+                createTrialWithMultipleLocations("Recruiting", "Belgium", "Recruiting", "Germany", "Active, not recruiting");
+        ClinicalTrial trialPotentiallyOpenInBelgiumSuspendedInGermany =
+                createTrialWithMultipleLocations("Recruiting", "Belgium", "Recruiting", "Germany", "Suspended");
+        ClinicalTrial trialPotentiallyOpenInBelgiumSuspendedInIndia =
```
kduyvesteyn, line 62: I think your test becomes more comprehensive when you make "suspendedInBelgiumOpenInIndia"?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        return CkbTestFactory.createEntry("BRAF", "BRAF V600E", "BRAF V600E", "sensitive", "Actionable", "AB", "AB", "A", "DOID:162");
+    }
+
+    @NotNull
+    private static List<VariantRequirementDetail> extractVariantRequirementDetails(@NotNull String requirementType) {
+        ClinicalTrial trial = CkbTestFactory.createTrialWithRequirementType(Lists.newArrayList(ImmutableVariantRequirementDetail.builder()
+                .profileId(0)
+                .requirementType(requirementType)
+                .build()));
+        return trial.variantRequirementDetails();
```
kduyvesteyn, line 103: There are various simplifications you can make here:
 - You first create VariantRequirementDetail in line 99, then create a ClinicalTrial, then return the variant requirement details anymore. The creation of the ClinicalTrial is unnecessary.
 - You have quite some duplicated code in the top level test. You can replace this with a function "hasRequirementTypeToInclude(string)", by expanding the createBasicEntry function

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+    @Test
+    public void canCreateActionableTrial() {
+        Location location = ImmutableLocation.builder().nctId("nctid").city("").country("Netherlands").status("Recruiting").build();
+        VariantRequirementDetail requirementType =
+                ImmutableVariantRequirementDetail.builder().profileId(0).requirementType("required").build();
+        CkbEntry entry = CkbTestFactory.createEntryWithClinicalTrial(location, "Recruiting", requirementType);
+        ActionableTrialFactory actionableTrialFactory = new ActionableTrialFactory();
+        Set<ActionableEntry> entrySet = actionableTrialFactory.create(entry, "KRAS", "gene");
+
+        assertEquals(1, entrySet.size());
```
kduyvesteyn, line 36: I haven't tested this but suspect your logic only works because both the variant requirement detail and entry have profileId 0. This is explicitly set in requirementType but not in entry, so this is an unwanted hidden dependency.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -0,0 +1,128 @@
...
+        assertEquals(Knowledgebase.CKB_TRIAL, trial.source());
+        assertEquals("KRAS", trial.sourceEvent());
+        assertEquals(Sets.newHashSet("https://clinicaltrials.gov/study/nctid"), trial.sourceUrls());
+        assertEquals("title", trial.treatment().name());
+        assertEquals("AB", trial.applicableCancerType().name());
+        assertEquals("162", trial.applicableCancerType().doid());
+        assertEquals(Sets.newHashSet(), trial.blacklistCancerTypes());
+        assertEquals(EvidenceLevel.B, trial.level());
+        assertEquals(EvidenceDirection.RESPONSIVE, trial.direction());
+        assertEquals(Sets.newHashSet("Netherlands"), trial.evidenceUrls());
```
kduyvesteyn, line 47: This test contains a fair amount of dependencies on CkbTestFactory which is unwanted. If you don't explicitly set a value when creating the CkbEntry you also shouldn't test for it afterwards. E.g. nowhere you set the applicable cancer type name to "AB" yet you test for it.

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
kduyvesteyn, line 1: I feel this test in general can be stronger. E.g. one test that is missing is that you don't create an ActionableEntry for an open trial in an allowed country when the variant requirement is on a different profile. I would at least add a test for that.

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
cbruel: Made two separate functions for these and made a separate function to create the ckbEntries

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
cbruel: Agree, added this test

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
cbruel: I changed this to set it explicitly in both

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
cbruel: Added the test

kduyvesteyn (approved):
Your changes all look good!

Some additional notes:
 - I made a bunch of changes which remove some unnecessary variables and make a more clear cut between CkbTestFactory and the individual test classes. Best to check ActionableTrialFactoryTest -> It tests exactly the thing you tested before but with less code/variables. 
 - `@NotNull Integer` is identical to `int` -> int is a base object and can by-design never be null. If you want to have an int that can be null, in java you are supposed to use `@Nullable Integer`.
 - Lists don't need to be initialised when constructing an immutable object from a builder. (e.g. `ageGroups(List.of())` is redundant.

PR #10, opened 2024-02-05
INFRA-236 introduce git tag release style

KasperWolsink99:
- Add `cloudbuild.yaml` that enables the git tag release style which we use (almost) everywhere now.
- Transition to monorepo style versioning where each module has the same version.
- Modify pom files to perform jooq code generation from DDL files instead of from a live database, allowing the builds to be cloudbuild friendly.


cloudbuild.yaml
```diff
@@ -0,0 +1,46 @@
...
+    args: [ 'versions:set', '-DnewVersion=${TAG_NAME}' ]
+    volumes:
+      - path: '/cache/.m2'
+        name: 'm2_cache'
+    env:
+      - MAVEN_OPTS=-Dmaven.repo.local=/cache/.m2
+  - id: 'Compile, package, release'
+    name: 'maven:3.9.2-eclipse-temurin-11'
+    entrypoint: mvn
+    args: [ 'deploy', '-B', '-DdeployAtEnd=true' ]
```
mvanniekerkHartwig, line 26: nice trick with `DdeployAtEnd`!

pom.xml
```diff
@@ -116,32 +116,6 @@
 
     <dependencyManagement>
         <dependencies>
-            <dependency>
```
mvanniekerkHartwig, line 119: Why not keep these with `${project.version}`. Than you can remove all of the versions in the modules.

mvanniekerkHartwig (commented):
Looks good, small question/suggestion.
Also, did you test the artifacts in artifact registry? Also, kind of part of this ticket was creating a docker image for Serve, but it's technically not a part of it. But I'll create another ticket to do that and I'll assign that to you.

pom.xml
KasperWolsink99: No particular reason, I think Mathijs recommended this way in another repo with multiple modules. Maybe it is more standard? Anyway I have incooporated the way you suggested as it does seem to be more inline with how we do it elsewhere.

KasperWolsink99:
> Looks good, small question/suggestion. Also, did you test the artifacts in artifact registry? Also, kind of part of this ticket was creating a docker image for Serve, but it's technically not a part of it. But I'll create another ticket to do that and I'll assign that to you.

I did test pulling in the artifacts as dependencies and this seems to work correctly

pom.xml
mvanniekerkHartwig: Ok, now you can get rid of `<version>${project.version}</version>` in all submodules.

KasperWolsink99: It should already be gone, it was included in the commit

mvanniekerkHartwig (approved):

PR #16, opened 2024-04-09
ACTIN-832: Configure FGFR2 dels as full exonic deletions

kduyvesteyn:


kzuberihmf (approved):

LiekeSchoenmaker (changes requested):
The CkbClassificationConfig should be updated to interpret the new FGFR events as fusion. See below the current interpretations of the events

LiekeSchoenmaker (approved):

kduyvesteyn:
> The CkbClassificationConfig should be updated to interpret the new FGFR events as fusion. Check the eventInterpretation for the current interpretations of the events

Thanks for double-checking! I've checked this, and it seems as-expected (del exon leads to a fusion, while exon del leads to an exon event)

PR #18, opened 2024-04-17
OSP-80: ckb trials oncoact support

LiekeSchoenmaker:
SCOPE: 
    - Optional for blacklisting clinical studies and evidence
    - Put clinical trial information into own model instead of using treatment model


kduyvesteyn (changes requested):
A few suggestions and high-level observations:
 - The ticket under which this work is done is OSP-80 and not OPR-80! 
 - For the datamodel change I suggest to make one interface that is extended by both Treatment and ClinicalTrial, and make that property `@NotNull`. 
 - I would harmonise "CKB Filtering" with "CKB Blacklisting". One idea would be to simply rename "CKB Filter" to "CKB Blacklist" which is top-level blacklisting (as opposed to on the level of evidence or studies). 

Finally, note that the datamodel changes are potentially non-backwards compatible in which case this will become v4.0.0 release.

PR #22, opened 2024-05-07
OSP-81: Filter the only children studies

LiekeSchoenmaker:


pauldwolfe (approved):

kzuberihmf (approved):
Looks good to me.

kduyvesteyn (commented):
These changes break compilation since `CkbExtractorTest` does not work anymore. The reason is that the test trials have no age groups defined. Can you fix this?

PR #26, opened 2024-05-15
ACTIN-953: Only annotate events when generateKnownEvents is set

jbartletthmf:
Replace synchronized list with map and collect.
Refactor using streaming conventions.


algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -130,174 +133,133 @@ private ExtractionResult getExtractionResult(CkbEntry entry, String gene, String
                 .interpretedEventType(entry.type())
                 .build();
 
-        ExtractionResult extraction = toExtractionResult(event, gene, null, extractionOutput, actionableEntries, interpretation);
-
-        if (generateKnownEvents) {
```
pauldwolfe, line 135: Where was the bug? Looks like the flag was checked?

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -85,33 +81,40 @@ public CkbExtractor(@NotNull Knowledgebase source, @NotNull final EventExtractor
...
+        String event = variantCount > 1 ? concat(entry.variants()) : CkbEventAndGeneExtractor.extractEvent(variant);
+        String gene = variantCount > 1 ? "Multiple" : CkbEventAndGeneExtractor.extractGene(variant);
+        
+        Optional<ExtractionResult> result;
+        if (entry.type() == EventType.UNKNOWN) {
+            LOGGER.warn("No event type known for '{}' on '{}'", event, gene);
+            result = Optional.empty();
+        } else {
+            ExtractionResult extraction = extractAndInterpretEvent(entry, gene, event);
+            result = Optional.of(generateKnownEvents ? CkbVariantAnnotator.annotate(extraction, variant) : extraction);
```
pauldwolfe, line 110: If we inlined the `extractAndInterpretEvent` maybe we can save some processing when flag is false and remove the second conditional checking generateKnownEvents?

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -130,174 +133,133 @@ private ExtractionResult getExtractionResult(CkbEntry entry, String gene, String
...
+                    a.addAllActionableCodons(built.actionableCodons());
+                    a.addAllActionableExons(built.actionableExons());
+                    a.addAllActionableGenes(built.actionableGenes());
+                    a.addAllActionableFusions(built.actionableFusions());
+                    a.addAllActionableCharacteristics(built.actionableCharacteristics());
+                    a.addAllActionableHLA(built.actionableHLA());
+                    return a;
+                });
 
         if (generateKnownEvents) {
```
pauldwolfe, line 183: this is the second conditional I was referring to.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
jbartletthmf: I extracted that function just because I thought this one was getting a bit long, but happy to refactor. I will say that in general I think we could find more efficient ways to accomplish CKB extraction as there are many places where we replace certain fields with new information that could probably be streamlined. For example, see `CkbVariantAnnotator.annotate` and `ExtractionFunctions.merge`.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
jbartletthmf: I'm not sure that inlining is the way to go because this is actually 2 levels deep (`getExtractionResult -> extractAndInterpretEvent -> toExtractionResult`). However, I don't see any reason why we can't inline the annotation at this level and skip the post-extraction annotation step.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
pauldwolfe: My point was that `extractAndInterpretEvent` is only called in this one place, and it is the only caller of `toExtractionResult`. So if we only call it when `generateKnownEvents` is true, then you don't need to check that whether `generateKnownEvents` is true on line 183.

jbartletthmf: `extractAndInterpretEvent` is called for every case where `entry.type() != EventType.UNKNOWN`, so we can't make any assumptions about `generateKnownEvents`.

pauldwolfe: I'm a bit confused, looking at the latest changes I think we missed each other. I meant to inline the call to `extractAndInterpretEvent` such that it would only be called if and when the generateKnownEvents was true. I guess I should have said "inline the variable `extraction`" and not the method... 

That said, if you like the way it looks now I think it still solves my concern, the flag is only checked in one place.

pauldwolfe (approved):

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
jbartletthmf: I still don't follow your suggestion...

We needed to call `extractAndInterpretEvent` when `generateKnownEvents` was false too. In that case, the `ExtractionResult` was returned without annotation. That said, annotation now happens with extraction so the extra check is no longer needed.

pauldwolfe: I see it now, I'd erroneously read the code as returning Optional.empty when the flag was false. My bad! I think its still nice now though that we don't need to check the flag twice. Alls well that ends well, sorry for any wasted time.

jbartletthmf: No problem, I still think this was an improvement.

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbVariantAnnotatorTest.java
```diff
@@ -27,40 +34,78 @@ public class CkbVariantAnnotatorTest {
...
+    private static ImmutableVariant variantWith(@NotNull Gene gene, @Nullable String proteinEffect) {
         return ImmutableVariant.builder()
                 .id(0)
                 .createDate(TEST_DATE)
                 .updateDate(TEST_DATE)
                 .fullName("variant")
                 .variant("variant")
                 .isHotspot(false)
-                .gene(gene)
+                .gene(gene).proteinEffect(proteinEffect).associatedWithDrugResistance("Y")
```
kduyvesteyn, line 108: Not a comment but just a remark: Nearly all actual variants have `associatedWithDrugResistance` null, seeing `Y` as default looks weird to me but I guess a test doesn't have to reflect reality :)

Also - FYI: There is no `N` value here, since CKB can never definitively know whether a variant implies drug resistance for any of the existing or future drugs!

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -85,255 +82,213 @@ public CkbExtractor(@NotNull Knowledgebase source, @NotNull final EventExtractor
...
+
+            EventInterpretation interpretation = ImmutableEventInterpretation.builder()
+                    .source(source)
+                    .sourceEvent(sourceEvent)
+                    .interpretedGene(gene)
+                    .interpretedEvent(event)
+                    .interpretedEventType(entry.type())
+                    .build();
+
+            List<CodonAnnotation> codons = curateCodons(extractionOutput.codons());
```
kduyvesteyn, line 122: Unrelated to your change, but maybe `curateCodons` should take in and return an `ExtractionOutput` such that we don't have to pass these specific codons around anymore.

Also, FYI: We curate codon 600 of BRAF because at that time the 38 ensembl cache used the wrong transcript for BRAF. I assume this is fixed by now (otherwise we'd have problems using OncoPanel 38!), we'd know for sure if we run a BRAF V600E through PAVE on 38 (or alternatively, run COLO pipeline on 38 since COLO has V600E mutation)

kduyvesteyn (approved):
Looks all good to me, some general comments:
 - `CkbExtractor` doesn't adhere to `serve_code_scheme`. This code scheme should be identical to ACTIN, so you should be able to auto-reformat.
 - You missed notnull annotations at various places. The policy is to annotate every input and output of every function + every property of a class. 
 - I'd suggest doing a regression on this upgrade just to be sure there is no actual change in SERVE output.

kduyvesteyn:
@pauldwolfe it seems like your cloud build is trying to publish a docker image upon each push and fails because it expects a tag? Shouldn't we only publish docker images upon the push of a tag?

pauldwolfe:
> @pauldwolfe it seems like the cloud build is trying to publish a docker image upon each push and fails because it expects a tag? Shouldn't we only publish docker images upon the push of a tag?

I'll have to make a separate PR cloudbuild.yaml that doesn't do the docker stuff. I'll do that this morning.

jbartletthmf:
IntelliJ made the formatting worse when I allowed reformatting before the commit, but calling Reformat directly seemed to fix it.

I'm trying to add all the annotations but I'm spoiled by Kotlin 😛

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -128,7 +128,8 @@ private ExtractionResult getExtractionResult(@NotNull CkbEntry entry) {
 
             if (generateKnownEvents) {
                 extractionResultBuilder.knownHotspots(convertToKnownHotspots(extractionOutput.hotspots(), event, variant))
-                        .knownCodons(convertToKnownCodons(extractionOutput.codons(), variant))
+                        .knownCodons(convertToKnownCodons(actionableEntries.isEmpty() ? Collections.emptyList() : extractionOutput.codons(),
+                                variant))
```
jbartletthmf, line 132: This change makes the output match the current research output, but is it desirable? In master, the input list of codons is initialized for each entry of `actionableEntries` or empty if there are no entries. Omitting input codons when there are no actionable entries reduces the known codon list from 124 to 30 (e.g. 6 codons for gene ERBB2, which is omitted entirely with this change).

ckb-importer/src/main/java/com/hartwig/serve/ckb/json/CkbJsonDirectoryReader.java
```diff
@@ -29,24 +32,28 @@ public CkbJsonDirectoryReader(@Nullable final Integer maxFilesToRead) {
 
     @NotNull
     public List<T> read(@NotNull String dir) throws IOException {
```
jbartletthmf, line 34: Parallelizing sped up reading from about 3 minutes to 1 minute on my laptop, which is significant when skipping hotspot resolution.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
kduyvesteyn: This means that we only consider a codon to be a known codon if the codon has associated clinical evidence? 

I think we've made that choice since CKB is actionability-focused, and we only want to consider a codon to be known (=pathogenic) if its evidence passes our filtering.

jbartletthmf: Thanks for confirming!

jbartletthmf:
@pauldwolfe not much has changed here since you gave approval so I'll merge this once I have a successful validation run if you have no objections.

pauldwolfe:
> @pauldwolfe not much has changed here since you gave approval so I'll merge this once I have a successful validation run if you have no objections.

No objections.

PR #28, opened 2024-05-21
OSP-84: Update the open recruitment types to include possible studies 

LiekeSchoenmaker:
I have discussed the different recruitment types to include possible studies for interpretation. For this reason a small update is necessary to inlcude the following recruitment types: 

Recruiting
Not yet recruiting
Available
Approved for marketing
not_yet_recruiting --> It is for me not clear why we should interpret this? 

see ticket https://hartwigmedical.atlassian.net/browse/OSP-84

### DIFF
--- a/algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
+++ b/algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
@@ -25,8 +25,13 @@ import org.jetbrains.annotations.Nullable;
 class ActionableTrialFactory implements ActionableEntryFactory {
 
     private static final String SUB_FIELD_DELIMITER = ",";
-    private static final Set<String> POTENTIALLY_OPEN_RECRUITMENT_TYPES =
-            Set.of("recruiting", "not yet recruiting", "not_yet_recruiting", "approved for marketing", "available");
+    private static final Set<String> POTENTIALLY_OPEN_RECRUITMENT_TYPES = Set.of("recruiting",
+            "not yet recruiting",
+            "not_yet_recruiting",
+            "approved for marketing",
+            "available",
+            "enrolling by invitation",
+            "enrolling_by_invitation");
     private static final Set<String> COUNTRIES_TO_INCLUDE = Set.of("netherlands", "belgium", "germany");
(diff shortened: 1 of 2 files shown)


pauldwolfe (changes requested):
We still require a ticket even for small changes.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -30,11 +31,9 @@ class ActionableTrialFactory implements ActionableEntryFactory {
 
     private static final Set<String> AGE_GROUPS_TO_INCLUDE = Sets.newHashSet();
 
-
     static {
         POTENTIALLY_OPEN_RECRUITMENT_TYPES.add("recruiting");
```
pauldwolfe, line 35: you can also create a set with `Set.of()` and remove the need for a mutable set and the static block

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -163,4 +162,9 @@ static boolean hasAgeGroupToInclude(@NotNull List<String> ageGroups) {
         }
         return false;
     }
+
+    @VisibleForTesting
+    static boolean hasPotentiallyOpenRequirementToInclude(@Nullable String locationStatus) {
```
pauldwolfe, line 167: Is `locationStatus` a really descriptive name? In one place the method is called with `trial.recruitment()`, and the values look more like a recruitment status than the status of a location.

pauldwolfe (approved):
Couple small ideas to improve, but approved. 

When you merge, please include the ticket in the title of the commit.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
LiekeSchoenmaker: done

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
LiekeSchoenmaker: indeed, I have update it to recruitmentStatus

cbruel (approved):

PR #29, opened 2024-05-21
OSP-83: Extend clinical trial model with gender and acronym

LiekeSchoenmaker:
Extend the clinical data model with gender and acronym. The study title and acronym is present into 2 separate fields, where acronym could be nullable


algo/src/main/resources/database/generate_serve_db.sql
```diff
@@ -14,6 +14,8 @@ CREATE TABLE `actionableHotspot`
     `sourceUrls` varchar(2000),
     `studyNctId` varchar(100),
     `studyTitle` varchar(500),
+    `studyAcronym` varchar(500),
```
kzuberihmf, line 17: Minor: I'd assume the acronym would be much shorter, so varchar(500) could be much more than required. But if not sure of length standards for the acronym then this is safer as is.

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtil.java
```diff
@@ -162,6 +166,8 @@ public static String toLine(@NotNull ActionableEvent event) {
                 .add(setToField(event.sourceUrls()))
                 .add(clinicalTrial != null ? clinicalTrial.studyNctId() : Strings.EMPTY)
                 .add(clinicalTrial != null ? clinicalTrial.studyTitle() : Strings.EMPTY)
+                .add(clinicalTrial != null ? clinicalTrial.studyAcronym() : Strings.EMPTY)
+                .add(clinicalTrial != null ? clinicalTrial.gender() : Strings.EMPTY)
```
kzuberihmf, line 170: Since the new fields are nullable, should they be set to null when no trial? I'm not sure what the conventions are.

LiekeSchoenmaker: Reason is that nullable values should be present as empty strings in the actionable output files

algo/src/main/java/com/hartwig/serve/sources/iclusion/ActionableTrialFactory.java
```diff
@@ -19,6 +19,7 @@
 import org.apache.commons.compress.utils.Lists;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
+import org.apache.logging.log4j.util.Strings;
```
jbartletthmf, line 22: I know this is used in a lot of places in our code, but this class is annotated `@InternalApi` with the description "Consider this class private": https://logging.apache.org/log4j/2.x/javadoc/log4j-api/org/apache/logging/log4j/util/Strings.html

I think it's better to just use `""` for the empty string instead of a potentially unstable dependency.

jbartletthmf (approved):
No major concerns 👍

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtil.java
kzuberihmf: Oh this was file writing, yes thanks!

algo/src/main/resources/database/generate_serve_db.sql
LiekeSchoenmaker: I think the length of this field can be indeed chorter. It is the acronym of the studyTitle. I have update it to varchar(250) becuae I am not sure what the required minimal varchar should be

PR #30, opened 2024-05-22
OSP-85: Correct interpretation of the treatment approach

LiekeSchoenmaker:
An incorrect interpretation of the treatment approach was made. After Discussion with CKB the correct interpretation should be: 
1.  TreatmentApporach  JSON and the treatmentApproach pages in BOOST will link together a specific Molecular Profile and either a drugClass or a Therapy. 
- If a treatmentApproach contains a drugClass then the therapy will be null, 
- If a treatmentApproach contains a therapy then the drugClass will be null, 
-->  They can have only one or the other.  Both BOOST and the JSON don’t specifically list any Efficacy evidence data on with the ProfileTreatmentApproach pages or JSON documents.

For implementation: 
1. Both fields therapy and drugClass are now interpret of the treatmentApproach model 
2. Both fields therapy and drugClass are now interpret for the high level evidence
--> Because none interpretation if the treatment approach field is currently done, this code will temporarily not used. This code will be used when corretly implemented


algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -100,43 +99,52 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
...
 
-                    String treatmentApproachString = String.join(",", sourceRelevantTreatmentApproaches);
-                    String treatmentApproachInterpret;
-                    if (sourceRelevantTreatmentApproaches.isEmpty()) {
-                        treatmentApproachInterpret = null;
-                    } else if (treatmentApproachString.endsWith(",")) {
-                        treatmentApproachInterpret = treatmentApproachString.substring(0, treatmentApproachString.length() - 1);
-                    } else {
-                        treatmentApproachInterpret = treatmentApproachString;
+                        if (therapyInfo != null) {
```
jbartletthmf, line 114: Maybe `else if` would be more appropriate here if these are assumed to be mutually exclusive.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -100,43 +99,52 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
...
+//                    TreatmentApproachCurationEntryKey key = ImmutableTreatmentApproachCurationEntryKey.builder()
+//                            .treatment(treatment)
+//                            .treatmentApproach(treatmentApproachInterpret == null || treatmentApproachInterpret.isEmpty()
+//                                    ? null
+//                                    : treatmentApproachInterpret)
+//                            .event(sourceGene + " " + entry.type())
+//                            .direction(direction)
+//                            .build();
+//
+//                    Set<String> curatedRelevantTreatmentApproaches = Sets.newHashSet(curator.isMatch(key));
```
jbartletthmf, line 138: Is this still relevant? Maybe it can be deleted.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -100,43 +99,52 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
                     sourceUrls.add("https://ckbhome.jax.org/profileResponse/advancedEvidenceFind?molecularProfileId=" + entry.profileId());
 
                     Set<String> sourceRelevantTreatmentApproaches = Sets.newHashSet();
```
kzuberihmf, line 101: Not used anymore?

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproachesFactory.java
```diff
@@ -36,8 +35,8 @@ private static RelevantTreatmentApproaches resolveRelevantTreatmentApproaches(@N
             if (treatmentApproach.id() == treatmentApproachInfo.id()) {
                 return ImmutableRelevantTreatmentApproaches.builder()
                         .id(treatmentApproach.id())
-                        .drugClass(treatmentApproach.drugClass() != null ? DrugFactory.resolveDrugClass(ckbJsonDatabase,
-                                treatmentApproach.drugClass()) : null)
+                        .drugClass(treatmentApproach.drugClass() != null ? treatmentApproach.drugClass() : null)
+                        .therapy(treatmentApproach.therapy() != null ? treatmentApproach.therapy() : null)
```
jbartletthmf, line 39: The ternary operator is redundant here:
```suggestion
                        .drugClass(treatmentApproach.drugClass())
                        .therapy(treatmentApproach.therapy())
```

ckb-importer/src/main/java/com/hartwig/serve/ckb/dao/TreatmentApproachDAO.java
```diff
@@ -42,20 +44,32 @@ public int write(@NotNull RelevantTreatmentApproaches treatmentApproaches) {
...
                         Tables.TREATMENTAPPROACHDRUGCLASS.TREATMENTAPPROACHID,
                         Tables.TREATMENTAPPROACHDRUGCLASS.DRUGCLASSID,
-                        Tables.TREATMENTAPPROACHDRUGCLASS.CREATEDATE,
                         Tables.TREATMENTAPPROACHDRUGCLASS.DRUGCLASS)
-                .values(treatmentApproachDrugClassId, drugClass.id(), drugClass.createDate(), drugClass.drugClass())
+                .values(treatmentApproachDrugClassId, drugClassInfo.id(), drugClassInfo.drugClass())
+                .execute();
+    }
+
+    private void writeTreatmentTherapy(@Nullable TherapyInfo therapyInfo, int treatmentApproachDrugClassId) {
```
kzuberihmf, line 67: Can this be @NotNull? Otherwise technically you need to check before extracting .id() etc. I think the only call site does a null check.

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproaches.java
```diff
@@ -17,7 +18,10 @@ public abstract class RelevantTreatmentApproaches {
     public abstract int id();
 
     @Nullable
-    public abstract DrugClass drugClass();
+    public abstract DrugClassInfo drugClass();
```
pauldwolfe, line 21: Having two nullable fields here means there's nothing in the code stopping someone setting both non-null.  Also, clients to this object can't make assumptions about whether its a drug based or therapy based, they need to always check.  And the alternative is possible, they could both be null. Does that make any sense? 

An alternative would be to have two classes DrugTreatmentApproach and TherapyTreatmentApproach, each with the appropriate "approach". They could extend a common superclass to reduce duplication. 

ActionableEvidenceFactory could then traverse both collections independently and without the nullable conditionals. 

I have not completely investigated the feasibility of doing this, maybe there is some context I'm missing. So please take as a suggestion for an improvement worth considering, not a mandatory change for approval.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -100,43 +99,52 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
...
+                    Set<String> drugClass = Sets.newHashSet();
```
pauldwolfe, line 102: These variable names should be plural, `drugClasses` and `therapies`

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
LiekeSchoenmaker: I have split it into drugClass and therapy set. So removed this unused set!

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
LiekeSchoenmaker: I have leave as is because this code is annotated as TODO. This should be correctly implemented in another round

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
LiekeSchoenmaker: I have annotated it with a TODO for implementation. This logica should be correctly implemented in another round so this cannot be deleted at this moment. However, this code will not be used at the moment.

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproachesFactory.java
LiekeSchoenmaker: implemented your suggestion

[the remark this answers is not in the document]
LiekeSchoenmaker: done

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
LiekeSchoenmaker: done

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproaches.java
LiekeSchoenmaker: I have implement some logics with a interface `TreatmentApproachIntervation` which is an optional intervation. Some context: 
- DrugClassTreatmentApproach contains data and TherapyTreatmentApproach is nullable
- TherapyTreatmentApproach contains data and DrugClassTreatmentApproach is nullable
- both DrugClassTreatmentApproach and TherapyTreatmentApproach are nullabel

pauldwolfe: Hey Lieke, not exactly what I had in mind for a couple of reasons:
- Empty marker interfaces should generally be avoided, they don't really have much use as you nearly always end up having to use instanceof/cast to figure out what they are.
- We still end up needing multiple checks on the nullability of the field to decide whether the approach is drug or therapy based.

I've created a new PR on top of this PR to illustrate: https://github.com/hartwigmedical/serve/pull/32. If you like it, you can merge that PR into this one.

LiekeSchoenmaker: thanks for your suggestion 👍 You have made a @NotNull list instead of a nullable item. I will merge it to the main OSP-85 branch

pauldwolfe (approved):

jbartletthmf (approved):
Is the `TreatmentApproachCurator` still needed? I assume it will be used when the `TODO` in the `ActionableEvidenceFactory` is addressed, but it should be cleaned up if not.

LiekeSchoenmaker:
The TreatmentApproachCurator will be used again when the TODO will be addressed in the next round. I prefer to leave as is

PR #31, opened 2024-05-22
ACTIN-975: Improve performance of hotspot extraction

jbartletthmf:
Parallelize DOCM and Hartwig hotspot extraction to allow concurrent transvar calls.
Add in-memory cache for Transvar for fast retrieval of previous queries.
Extract CKB entries once and share with evidence and trial extraction.
Streamline ServeAlgo and RefGenomeManager using streams.


algo/src/main/java/com/hartwig/serve/ServeAlgo.java
```diff
@@ -71,59 +71,28 @@ public ServeAlgo(@NotNull final RefGenomeManager refGenomeManager, @NotNull fina
...
-        }
-
-        Map<RefGenome, List<ExtractionResult>> versionedMap = refGenomeManager.makeVersioned(extractions);
-
-        Map<RefGenome, ExtractionResult> refDependentExtractionMap = Maps.newHashMap();
-        for (Map.Entry<RefGenome, List<ExtractionResult>> entry : versionedMap.entrySet()) {
-            refDependentExtractionMap.put(entry.getKey(), ExtractionFunctions.merge(entry.getValue()));
-        }
+        Stream<ExtractionResult> ckbExtractions =
+                (config.useCkbEvidence() || config.useCkbTrials()) ? extractCkbKnowledge(config) : Stream.empty();
```
pauldwolfe, line 75: Could we combine this with main config to extraction stream somehow? I think even in it incurs a bit of duplication it will be a little more readable.

pauldwolfe (approved):

algo/src/main/java/com/hartwig/serve/ServeAlgo.java
jbartletthmf: Is the latest change what you had in mind? The main thing I wanted to achieve here was only reading the CKB entries once.

pauldwolfe: Got it, yeah what you did in the last commit looks good to me!

kzuberihmf (approved):
Cool. Wasn't sure how the parallelism was controlled, but apparently it defaults to the # of cpu cores available? That seems reasonable, I think it means that if parallelizing the transvar calls, the mem requirements will be +N*mem(transvar-python), each of which read annotations & possibly sequence. Likely not a problem, but something to keep in mind operationally. 

Are you able to comment on the approx runtime performance change you observed?

jbartletthmf:
FYI the approach of parallelizing transvar calls isn't new here, I just propagated it from the CkbExtractor (which has the longest runtime) to some of the smaller extraction jobs. As for runtime changes:

1. I parallelized CKB entry reading in my other PR, which sped it up from 2-3 minutes to about 12s on the VM. Reading once and reusing thus saves about 12s.
2. Extracting the CKB data takes about 42 minutes, mainly due to transvar. Adding caching in transvar allows the second iteration (evidence and trials are extracted separately) to occur in less than a second, saving 42 minutes.
3. DoCM and Hartwig hotspot extraction now take about 1 minute each, when they previously took about 20 and 6 minutes, respectively. This represents a reduction of about 24 minutes.

The entire SERVE run went from 1h53m to 46 minutes, so I observed a reduction in runtime > 50%.

kzuberihmf:
> FYI the approach of parallelizing transvar calls isn't new here, I just propagated it from the CkbExtractor (which has the longest runtime) to some of the smaller extraction jobs. As for runtime changes:
> 
> 1. I parallelized CKB entry reading in my other PR, which sped it up from 2-3 minutes to about 12s on the VM. Reading once and reusing thus saves about 12s.
> 2. Extracting the CKB data takes about 42 minutes, mainly due to transvar. Adding caching in transvar allows the second iteration (evidence and trials are extracted separately) to occur in less than a second, saving 42 minutes.
> 3. DoCM and Hartwig hotspot extraction now take about 1 minute each, when they previously took about 20 and 6 minutes, respectively. This represents a reduction of about 24 minutes.
> 
> The entire SERVE run went from 1h53m to 46 minutes, so I observed a reduction in runtime > 50%.

Great info, thanks!

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/variant/VariantFactory.java
```diff
@@ -16,72 +18,74 @@
...
-                        .referenceTranscriptCoordinate(convertReferenceTranscriptCoordinate(variant.referenceTranscriptCoordinate()))
-                        .allTranscriptCoordinates(convertAllTranscriptCoordinates(variant.allTranscriptCoordinates()))
-                        .categoryVariantPaths(convertCategoryVariantPaths(variant.categoryVariantPaths()))
-                        .memberVariants(convertMemberVariants(variant.memberVariants()))
-                        .description(ReferenceFactory.extractDescription("variant", variant.id(), variant.descriptions()))
-                        .references(ReferenceFactory.extractDescriptionReferences(ckbJsonDatabase, variant.descriptions()))
-                        .build();
-            }
+    private Variant resolveVariant(@NotNull VariantInfo variantInfo) {
+        if (variantsById.containsKey(variantInfo.id())) {
```
LiekeSchoenmaker, line 41: should this not equals? Otherwise id of e.g. 10 versus 100 can be also true

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/variant/VariantFactory.java
```diff
@@ -16,72 +18,74 @@
...
-                        .mapLocation(gene.mapLocation())
-                        .canonicalTranscript(gene.canonicalTranscript())
-                        .terms(gene.terms())
-                        .synonyms(gene.synonyms())
-                        .description(ReferenceFactory.extractDescription("gene", gene.id(), gene.descriptions()))
-                        .references(ReferenceFactory.extractDescriptionReferences(ckbJsonDatabase, gene.descriptions()))
-                        .build();
-            }
+    private Gene resolveGene(@NotNull GeneInfo geneInfo) {
+        if (genesById.containsKey(geneInfo.id())) {
```
LiekeSchoenmaker, line 72: should this not equals? Otherwise id of e.g. 10 versus 100 can be also true

LiekeSchoenmaker:
I have put 2 small remarks. I am not very familiar with the hotspot extraction part of serve, but looks further good to me

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/variant/VariantFactory.java
jbartletthmf: `variantsById` is a `Map<Integer, JsonVariant>`, so [containsKey()](https://docs.oracle.com/javase/8/docs/api/java/util/Map.html#containsKey-java.lang.Object-) will check for the presence of that specific integer in the map's keyset (unlike `String.contains()`). I introduced a map here because they allow very fast random access for known keys, which means each variant can be resolved almost instantaneously. This is much more efficient than the previous approach of requiring a full scan of all variants in ckbJsonDatabase for every `variantInfo` that needs to be resolved.

PR #32, opened 2024-05-23
OSP-85 Suggested refactoring to simplify use of treatment approaches

pauldwolfe:
Does the following:
- Creates specific classes for drug and therapy based approaches
- Splits the collection into one drug collection and one therapy collection in the evidence.
- Moves the nullability check to one place, in the evidence creation when populating the two new collections.
- Use streams in the actionable evidence factory.

There is still some duplication that could be removed, I'll annotate it in the PR.

### DIFF
--- a/ckb-importer/src/main/java/com/hartwig/serve/ckb/dao/EvidenceDAO.java
+++ b/ckb-importer/src/main/java/com/hartwig/serve/ckb/dao/EvidenceDAO.java
@@ -6,7 +6,9 @@ import com.hartwig.serve.ckb.database.tables.Indicationevidence;
 import com.hartwig.serve.ckb.database.tables.Therapyevidence;
 import com.hartwig.serve.ckb.database.tables.Treatmentapproachevidence;
 import com.hartwig.serve.ckb.datamodel.reference.Reference;
+import com.hartwig.serve.ckb.datamodel.treatmentapproaches.DrugClassTreatmentApproach;
 import com.hartwig.serve.ckb.datamodel.treatmentapproaches.RelevantTreatmentApproaches;
+import com.hartwig.serve.ckb.datamodel.treatmentapproaches.TherapyTreatmentApproach;
 
 import org.jetbrains.annotations.NotNull;
 import org.jooq.DSLContext;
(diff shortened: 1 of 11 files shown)


ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproachesFactory.java
```diff
@@ -19,49 +19,58 @@ private RelevantTreatmentApproachesFactory() {
     }
 
     @NotNull
-    public static List<RelevantTreatmentApproaches> extractRelevantTreatmentApproaches(@NotNull CkbJsonDatabase ckbJsonDatabase,
+    public static List<DrugClassTreatmentApproach> extractDrugTreatmentApproaches(@NotNull CkbJsonDatabase ckbJsonDatabase,
```
pauldwolfe, line 22: There is a bunch of duplication here that could be refactored out between the two public methods.

ckb-importer/src/main/java/com/hartwig/serve/ckb/dao/EvidenceDAO.java
```diff
@@ -71,7 +73,17 @@ public void write(@NotNull com.hartwig.serve.ckb.datamodel.evidence.Evidence evi
                 Indicationevidence.INDICATIONEVIDENCE.EVIDENCEID,
                 Indicationevidence.INDICATIONEVIDENCE.INDICATIONID).values(id, indicationId).execute();
 
-        for (RelevantTreatmentApproaches treatmentApproaches : evidence.relevantTreatmentApproaches()) {
+        for (DrugClassTreatmentApproach treatmentApproaches : evidence.drugTreatmentApproaches()) {
```
pauldwolfe, line 76: This duplication could be refactored.

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproachesFactory.java
```diff
@@ -19,49 +19,58 @@ private RelevantTreatmentApproachesFactory() {
...
-                    .drugClass(DrugFactory.resolveDrugClass(ckbJsonDatabase, Objects.requireNonNull(treatmentApproach.drugClass())))
-                    .build();
-        } else {
-            throw new IllegalStateException("An treatment approach cannot be both a drug class and therapy!");
+    @NotNull
+    private static DrugClassTreatmentApproach resolveDrugTreatmentApproaches(@NotNull CkbJsonDatabase ckbJsonDatabase,
+            @NotNull TreatmentApproachInfo treatmentApproachInfo) {
+
+        for (JsonTreatmentApproach treatmentApproach : ckbJsonDatabase.treatmentApproaches()) {
+            if (treatmentApproach.id() == treatmentApproachInfo.id() && treatmentApproach.drugClass() != null) {
```
jbartletthmf, line 52: This is a common antipattern in SERVE where we do a full DB scan for each ID we need to resolve instead of building a map first. I fixed this in `VariantFactory` in https://github.com/hartwigmedical/serve/pull/31 by creating an instance of the factory that holds the maps, but the issue is so prevalent we may be better off converting the lists in the `CkbJsonDatabase` to maps so we can support fast lookups with static use of the factories. This is out of scope here, but just curious if anyone else has thoughts. Also, I recognize that these changes are unlikely to significantly impact SERVE's runtime for most use cases 🙂

jbartletthmf (approved):

ckb-importer/src/main/java/com/hartwig/serve/ckb/datamodel/treatmentapproaches/RelevantTreatmentApproachesFactory.java
LiekeSchoenmaker: I will leave this to another round or improvements

PR #35, opened 2024-07-08
ACTIN-1099 Add CKB region configuration to serve

pauldwolfe:


kduyvesteyn (changes requested):
I would do it a bit more cleanly, as follows:

Add a "country config tsv" input with 2 columns:
 - Country 
 - Set of states to include from country (comma separated or whatever is standard). 

The set of states to include could have a special value "all" in which case all states are included. Then for our current use-case we'd configure NL, Germany, Belgium with all states (though could even remove German states on the Eastern part of Germany potentially).

Also, I would put the param closer to the other CKB params and maybe even include "ckb" in the name of the param to make it clear that this param is used within the context of CKB (similar to `ckb_blacklist_trial_tsv`)

pauldwolfe:
> I would do it a bit more cleanly, as follows:
> 
> Add a "country config tsv" input with 2 columns:
> 
> * Country
> * Set of states to include from country (comma separated or whatever is standard).
> 
> The set of states to include could have a special value "all" in which case all states are included. Then for our current use-case we'd configure NL, Germany, Belgium with all states (though could even remove German states on the Eastern part of Germany potentially).
> 
> Also, I would put the param closer to the other CKB params and maybe even include "ckb" in the name of the param to make it clear that this param is used within the context of CKB (similar to `ckb_blacklist_trial_tsv`)

Sounds good to me! I had forgotten about the states.

kduyvesteyn:
> > I would do it a bit more cleanly, as follows:
> > Add a "country config tsv" input with 2 columns:
> > 
> > * Country
> > * Set of states to include from country (comma separated or whatever is standard).
> > 
> > The set of states to include could have a special value "all" in which case all states are included. Then for our current use-case we'd configure NL, Germany, Belgium with all states (though could even remove German states on the Eastern part of Germany potentially).
> > Also, I would put the param closer to the other CKB params and maybe even include "ckb" in the name of the param to make it clear that this param is used within the context of CKB (similar to `ckb_blacklist_trial_tsv`)
> 
> Sounds good to me! I had forgotten about the states.

Like the direction this is going! Have made a bunch of smallish changes and added one TODO in code (CkbRegion) where maybe something is going wrong (depending on final implementation though).

algo/src/main/java/com/hartwig/serve/sources/ckb/region/CkbRegion.java
```diff
@@ -0,0 +1,25 @@
...
+
+    @NotNull
+    public abstract String country();
+
+    @NotNull
+    public abstract List<String> states();
+
+    public boolean includes(@NotNull Location location) {
+        // TODO (PW): Shouldn't this be "states.isEmpty || states.contains()"? Or how would someone configure all states for a country?
+        return location.country().equals(country()) && (!states().isEmpty() && states().contains(location.state()));
```
pauldwolfe, line 23: I was thinking that empty states = all states. It seems logical to me if you only specify a country, you get everything for that country. Otherwise you need to both add a special string and some error handling for empty states (which then will match nothing).

pauldwolfe: That said, you're TODO was right that the logic was inverted :P

kduyvesteyn: I have no strong opinion, you could also make it null perhaps (null = all states) and rename the param to `specificStates`, though also fine with empty = all.

kduyvesteyn: I guess cleanest would be to have a predicate that is either alwaysTrue or filters on specific state :) but sounds like overkill

algo/src/main/java/com/hartwig/serve/ServeConfig.java
```diff
@@ -69,6 +70,7 @@ public interface ServeConfig {
     // Options to help with debugging / testing
     String SKIP_HOTSPOT_RESOLVING = "skip_hotspot_resolving";
     String LOG_DEBUG = "log_debug";
+    String COUNTRIES_TO_INCLUDE = "countries";
```
kduyvesteyn, line 73: Variable can be cleaned up!

algo/src/main/java/com/hartwig/serve/sources/ckb/region/CkbRegion.java
```diff
@@ -0,0 +1,40 @@
...
+public abstract class CkbRegion {
+
+    @NotNull
+    public abstract String country();
+
+    @NotNull
+    public abstract Set<String> states();
+
+    public boolean includes(@NotNull Location location) {
+        return country().equals(toLowercaseNullable(location.country())) && (states().isEmpty() || states().contains(toLowercaseNullable(
```
kduyvesteyn, line 24: Minor but I would do this in 2 lines (first line would be `boolean isStateMatch`

algo/src/main/java/com/hartwig/serve/sources/ckb/region/CkbRegion.java
```diff
@@ -0,0 +1,40 @@
...
+                location.state())));
+    }
+
+    @NotNull
+    public static CkbRegion of(@NotNull String country, @NotNull Set<String> states) {
+        return ImmutableCkbRegion.builder()
+                .country(country.toLowerCase())
+                .states(states.stream().map(String::toLowerCase).collect(Collectors.toSet()))
+                .build();
+    }
```
kduyvesteyn, line 34: Also minor but it's not typical to have this function. The production code generally only has one place to instantiate (where we would use the Immutable directly) and for test we'd use a test factory. If you want to leave this in place fine with me though

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -30,6 +33,13 @@
 public class ActionableTrialFactoryTest {
 
     private static final CkbStudyBlacklistModel BLACKLIST_MODEL = CkbBlacklistTestFactory.createProperStudyBlacklist();
+    public static final Set<CkbRegion> COUNTRIES_TO_INCLUDE =
```
kduyvesteyn, line 36: Would rename to `REGIONS_TO_INCLUDE`

algo/src/test/java/com/hartwig/serve/sources/ckb/region/CkbRegionFileTest.java
```diff
@@ -0,0 +1,36 @@
...
+
+public class CkbRegionFileTest {
+
+    @Test
+    public void shouldLoadFileWithAndWithoutStates() throws Exception {
+        Set<CkbRegion> regions = CkbRegionFile.read(Resources.getResource("ckb_region/ckb_regions.tsv").getPath());
+        CkbRegion netherlands = regions.stream().filter(r -> r.country().equals("netherlands")).findFirst().orElseThrow();
+        CkbRegion us = regions.stream().filter(r -> r.country().equals("united states")).findFirst().orElseThrow();
+
+        assertTrue(netherlands.includes(location("Netherlands", "Nord Holland")));
```
kduyvesteyn, line 23: Fairly minor but at the same time super important: Correct spelling is "Noord Holland" :)

algo/src/test/java/com/hartwig/serve/sources/ckb/region/CkbRegionFileTest.java
```diff
@@ -0,0 +1,36 @@
...
+        assertTrue(us.includes(location("United States", "Maine")));
+        assertFalse(us.includes(location("United States", "California")));
+        assertFalse(us.includes(location("United States", null)));
+
+    }
+
+    @NotNull
+    private static ImmutableLocation location(@NotNull String country, @Nullable String state) {
+        return ImmutableLocation.builder().country(country).state(state).city("city").nctId("nct").build();
+    }
```
kduyvesteyn, line 35: I think you can reuse `CkbTestFactory` here? Otherwise, should be able to return Location rather than ImmutableLocation

kduyvesteyn (approved):
Looks good! 

Added a few minor comments, plus added Lieke as reviewer

algo/src/main/java/com/hartwig/serve/ServeConfig.java
```diff
@@ -244,6 +250,7 @@ static ServeConfig createConfig(@NotNull CommandLine cmd) throws ParseException
                 .ckbBlacklistTrialTsv(useCkbTrials ? nonOptionalFile(cmd, CKB_BLACKLIST_TRIAL_TSV) : NOT_APPLICABLE)
                 .ckbBlacklistEvidenceTsv(useCkbEvidence ? nonOptionalFile(cmd, CKB_BLACKLIST_EVIDENCE_TSV) : NOT_APPLICABLE)
                 .ckbDrugCurationTsv(useCkbEvidence ? nonOptionalFile(cmd, CKB_DRUG_CURATION_TSV) : NOT_APPLICABLE)
+                .ckbRegionTsv(useCkbEvidence ? nonOptionalFile(cmd, CKB_REGION_TSV) : NOT_APPLICABLE)
```
LiekeSchoenmaker, line 253: should this not be useCkbTrials instead of useCkbEvidence?

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -119,10 +123,10 @@ && hasPotentiallyOpenRequirementToInclude(trial.recruitment()) && hasAgeGroupToI
 
     @NotNull
     @VisibleForTesting
-    static Set<String> countriesToInclude(@NotNull ClinicalTrial trial) {
+    static Set<String> filterOnRegionsToInclude(@NotNull ClinicalTrial trial, @NotNull Set<CkbRegion> regionsToInclude) {
         Set<String> countries = Sets.newHashSet();
         for (Location location : trial.locations()) {
-            if (COUNTRIES_TO_INCLUDE.contains(location.country().toLowerCase())
+            if (regionsToInclude.stream().anyMatch(region -> region.includes(location))
```
LiekeSchoenmaker, line 129: Should you not use location.region or something?

LiekeSchoenmaker (changes requested):
Good improvements to configure the countries in a TSV file instead hard-coded in the algorithm itself!

I have put some small requests. I think one is necessary for correct use of the country config

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
pauldwolfe: This is the way I've intended it. A region in our case is one country + a list of states (or provinces). A location is a specific location with a country and a state. You can check if a location is within a region with the includes method.

Regions are user defined at runtime so it doesn't really make sense to attach them to locations.

algo/src/main/java/com/hartwig/serve/sources/ckb/region/CkbRegion.java
pauldwolfe: agrees that is a readability improvement.

algo/src/main/java/com/hartwig/serve/sources/ckb/region/CkbRegion.java
pauldwolfe: Ok, this a pattern I normally use to allow clients of the immutable not to have to know they are using an immutable (ie depend only on the interface). But I think its most important to be consistent, so I'll move this elsewhere.

PR #36, opened 2024-08-05
ACTIN-1166: Add 'no benefit' and 'decreased response' to directions to output

cbruel:


algo/README.md
```diff
@@ -449,8 +450,10 @@ elsewhere.
 
 ## Version History and Download Links
 
+- Upcoming:
+    - Add "Decreased response" and "No benefit" to directions to output
```
kduyvesteyn, line 454: This is not an entirely correct sentence I think! Maybe just say "to evidence directions" and remove "to output"?

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -47,12 +49,13 @@ class ActionableEvidenceFactory implements ActionableEntryFactory {
...
         PREDICTED_RESISTANT_DIRECTIONS.add("predicted - resistant");
 
+        NO_BENEFIT_DIRECTIONS.add("no benefit");
+        DECREASED_RESPONSE_DIRECTIONS.add("decreased response");
+
         DIRECTIONS_TO_IGNORE.add("unknown");
         DIRECTIONS_TO_IGNORE.add("not applicable");
         DIRECTIONS_TO_IGNORE.add("conflicting");
-        DIRECTIONS_TO_IGNORE.add("no benefit");
         DIRECTIONS_TO_IGNORE.add("not predictive");
```
kduyvesteyn, line 58: Did you check if "not predictive" may be mappable to `NO_BENEFIT`?

datamodel/src/main/java/com/hartwig/serve/datamodel/EvidenceDirection.java
```diff
@@ -1,24 +1,31 @@
...
-        this.isResponsive = isResponsive;
+    EvidenceDirection(final boolean hasPositiveResponse, final boolean hasNoBenefit, final boolean isResistant, final boolean isCertain) {
+        this.hasPositiveResponse = hasPositiveResponse;
+        this.hasNoBenefit = hasNoBenefit;
         this.isResistant = isResistant;
         this.isCertain = isCertain;
     }
 
-    public boolean isResponsive() {
-        return isResponsive;
```
kduyvesteyn, line 21: You are not allowed to remove a function from a datamodel object since this will break backwards compatibility.

What you need to do is to keep this function, have it return `hasPositiveResponse` (assuming that's correct at least!), and annotate it with `@Deprecated`, so that clients know they should not use this function anymore.

datamodel/src/main/java/com/hartwig/serve/datamodel/EvidenceDirection.java
```diff
@@ -1,24 +1,31 @@
...
+    private final boolean hasPositiveResponse;
+    private final boolean hasNoBenefit;
     private final boolean isResistant;
     private final boolean isCertain;
 
-    EvidenceDirection(final boolean isResponsive, final boolean isResistant, final boolean isCertain) {
-        this.isResponsive = isResponsive;
+    EvidenceDirection(final boolean hasPositiveResponse, final boolean hasNoBenefit, final boolean isResistant, final boolean isCertain) {
+        this.hasPositiveResponse = hasPositiveResponse;
+        this.hasNoBenefit = hasNoBenefit;
```
kduyvesteyn, line 18: It is more typical to define booleans in terms of their "positive" intent, so this variable should then be called "hasBenefit". Or maybe a better name would be "impliesResponsivenessOrResistance`? or `impliesImpact`?

kduyvesteyn (commented):
Looks good! A bunch of smallish suggestions.

Also, it would be good to double-check if this upgrade can logically be done without updating ACTIN at the same time? (e.g. if ACTIN iterates over all potential directions then maybe that wouldn't work out well, or if ACTIN assumes that `isResponsive == !isResistant` then we need to update that as well prior to this SERVE upgrade.

kduyvesteyn:
One additional comment: Would remove `@SuppressWarnings("unused")` from line 35 in `EvidenceDirection`. It doesn't add much and is inconsistent with other datamodel functions that are unused in SERVE itself.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
cbruel: I looked at the documents in the weekly dump, and "not predictive" is not in there, so overlooked it but the definition is on CKB's website (I think "not predictive" is quite new, the docs in the weekly dump are from 2019..). Looking at the definition: _"Evidence from peer-reviewed published literature from one or more studies demonstrating that a molecular profile does not confer sensitivity or resistance to a specific monotherapy or combination therapy"_, I think we could indeed map "not predictive" to `NO_BENEFIT`.

datamodel/src/main/java/com/hartwig/serve/datamodel/EvidenceDirection.java
cbruel: I personally prefer `hasBenefit`. Some background on my logic for the booleans:
- `hasPositiveResponse` makes sure we can distinct "sensitive" (positive response) from "decreased response", "resistant", "no benefit" (negative response).
- Then `isResistant `makes a distinction between "resistant" (resistant) and "no benefit" & "decreased response" (no resistance).
- Lastly `hasBenefit` makes a distinction between "no benefit" (has no benefit) and "decreased response" (has benefit)

cbruel:
@kduyvesteyn Checked where and how this datamodel is used in actin, and I would say we can upgrade without issues

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
kduyvesteyn: There are actually >100 entries (see also `select * from actionability where responseType = 'not predictive';` in `ckb_latest`). I've scanned a bunch of them and they feel quite similar to `NO_BENEFIT`. Sounds like a good improvement to map these then! I will double-check with Sara nonetheless

datamodel/src/main/java/com/hartwig/serve/datamodel/EvidenceDirection.java
kduyvesteyn, line 1: I leave it up to you if you want to add this but I just manually checked the configuration of every boolean for every enum value. You could also do this in a unit test, to make sure no one accidentally changes one of the booleans. It is a bit of duplication so not sure if it is a good idea, either way is fine for me!

kduyvesteyn (approved):

datamodel/src/main/java/com/hartwig/serve/datamodel/EvidenceDirection.java
cbruel: I like the idea. I personally think testing a bit too much, is always better than too less

PR #37, opened 2024-08-09
ACTIN-1032: Add German/Belgian cities and dutch hospitals

cbruel:
See jira ticket for more information.

FYI. I tested extensively by:
- Ran my little algorithm (which determines the facility name) on all CKB locations we currently have for Dutch trials. And checked if the curated facility name was correct (correct for all 800+ locations).
- Ran `run_serve_pilot_ckb_trials_only` (first without hotspot resolving, later with) and checked the output files.


algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -42,10 +50,14 @@ class ActionableTrialFactory implements ActionableEntryFactory {
     private final CkbStudyBlacklistModel blacklistStudy;
     @NotNull
     private final Set<CkbRegion> regionsToInclude;
+    @NotNull
+    private final CkbFacilityCurationModel ckbFacilityCuration;
```
kduyvesteyn, line 54: No need to include `ckb` in the name here since it is implied by the context

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -54,7 +66,7 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
         Set<ActionableEntry> actionableTrials = Sets.newHashSet();
 
         for (ClinicalTrial trial : trialsToInclude(entry)) {
-            Set<String> countries = filterOnRegionsToInclude(trial, regionsToInclude);
+            Set<Country> countries = filterOnRegionsToInclude(trial, regionsToInclude, ckbFacilityCuration);
```
kduyvesteyn, line 69: Maybe rename this function to be more representative? e.g. `extractCountriesToInclude`?

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -123,16 +135,33 @@ && hasPotentiallyOpenRequirementToInclude(trial.recruitment()) && hasAgeGroupToI
...
+            @NotNull CkbFacilityCurationModel ckbFacilityCurationModel) {
+        Map<String, Map<String, Set<String>>> countriesToCitiesToHospitalNames = trial.locations()
+                .stream()
+                .filter(location -> regionsToInclude.stream().anyMatch(region -> region.includes(location))
+                        && hasPotentiallyOpenRequirementToInclude(location.status()))
+                .collect(Collectors.groupingBy(Location::country, Collectors.groupingBy(Location::city, Collectors.mapping(location -> {
+                    if ("Netherlands".equals(location.country())) {
+                        return ckbFacilityCurationModel.curateFacilityName(location);
+                    }
+                    return "";
```
kduyvesteyn, line 148: I would not make the curation dependent on "Netherlands", I would rather do pass-through in case there is no curation entry. Maybe AUS or USA want to use this curation as well and would be silly if that is not possible due to hard-coding!

algo/src/main/java/com/hartwig/serve/sources/ckb/facility_curation/CkbFacilityCurationFilterFile.java
```diff
@@ -0,0 +1,50 @@
...
+import java.nio.file.Files;
+import java.util.List;
+import java.util.Map;
+
+import com.hartwig.serve.datamodel.serialization.util.SerializationUtil;
+
+import org.apache.commons.compress.utils.Lists;
+import org.jetbrains.annotations.NotNull;
+
+public class CkbFacilityCurationFilterFile {
```
kduyvesteyn, line 14: Class with only static functions should be `final` so that people can't derive from it.

algo/src/main/java/com/hartwig/serve/sources/ckb/facility_curation/CkbFacilityCurationModel.java
```diff
@@ -0,0 +1,101 @@
...
+        }
+        if (string1 == null) {
+            return false;
+        }
+        return string1.equals(string2);
+    }
+
+    @VisibleForTesting
+    @NotNull
+    Boolean containsWord(@Nullable String string1, @NotNull String string2) {
```
kduyvesteyn, line 97: This can return a `boolean` I think since you never return `null`

algo/src/main/java/com/hartwig/serve/sources/ckb/facility_curation/CkbFacilityCurationModel.java
```diff
@@ -0,0 +1,101 @@
...
+                unusedFacilityCurationFilterCount++;
+                LOGGER.warn(" Facility curation filter entry '{}' hasn't been used for CKB filtering", entry);
+            }
+        }
+        LOGGER.debug(" Found {} unused facility curation filters during CKB filtering", unusedFacilityCurationFilterCount);
+    }
+
+    @VisibleForTesting
+    @NotNull
+    Boolean equalStringsOrNull(@Nullable String string1, @NotNull String string2) {
```
kduyvesteyn, line 85: This can return a `boolean` I think because you never return `null`

algo/src/main/java/com/hartwig/serve/sources/ckb/facility_curation/CkbFacilityCurationNameFile.java
```diff
@@ -0,0 +1,45 @@
...
+import java.nio.file.Files;
+import java.util.List;
+import java.util.Map;
+
+import com.hartwig.serve.datamodel.serialization.util.SerializationUtil;
+
+import org.apache.commons.compress.utils.Lists;
+import org.jetbrains.annotations.NotNull;
+
+public class CkbFacilityCurationNameFile {
```
kduyvesteyn, line 14: Can be final!

kduyvesteyn (changes requested):
Have only reviewed on a high level as I believe there are some datamodel changes that need some thought. 

Also, in general, the datamodel has no `dutchHospitals` while derived classes do, that should not be possible (and besides, the datamodel should not have a dependency on NL)

algo/src/main/java/com/hartwig/serve/sources/ckb/facility_curation/CkbFacilityCurationFilterEntry.java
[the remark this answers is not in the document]
cbruel: @kduyvesteyn the term "facility" comes from "facility name" field in CKB's datamodel, this is part of `select * from location`

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtil.java
```diff
@@ -36,11 +41,12 @@ public static String header() {
...
-                .add("studyNctId")
-                .add("studyTitle")
-                .add("studyAcronym")
-                .add("studyGender")
-                .add("countriesOfStudy")
+                .add("nctId")
+                .add("title")
+                .add("acronym")
+                .add("genderCriterium")
+                .add("countries")
```
kduyvesteyn, line 48: Is this field really just countries or actually "countries and cities"? Code seems to imply both

datamodel/src/main/java/com/hartwig/serve/datamodel/Country.java
```diff
@@ -0,0 +1,22 @@
...
+public abstract class Country {
+
+    @NotNull
+    public abstract String countryName();
+
+    @NotNull
+    public abstract Set<String> cities();
+
+    @Nullable
+    public abstract Map<String, Set<String>> hospitalsPerCity();
```
kduyvesteyn, line 21: Will every city always appear as a key in the "hospitals per city" map? 

If yes, you could remove the `cities` property since it is implied by the keyset of `hospitalsPerCity`

I assume no since `hospitalsPerCity` is nullable and maybe sometimes not filled in?

My feeling is that it would be cleaner to relate the 2 fields (every city will always be a key in the map, though the value could be empty (or even null))

datamodel/src/test/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtilTest.java
```diff
@@ -49,34 +53,70 @@ public void canConvertActionableEvents() {
         assertEquals(event.evidenceUrls(), coveredEvent.evidenceUrls());
     }
 
+    @Test
+    public void canConvertCountriesOfStudyToHospitals() {
+        Set<Country> countriesOfStudy = Sets.newHashSet();
+        countriesOfStudy.add(createCountry("Netherlands", "Groningen", "UMCG"));
+        countriesOfStudy.add(createCountry("Belgium", "Brussel", "UZ Brussel"));
+        assertEquals("Brussel(UZ Brussel),Groningen(UMCG)", ActionableFileUtil.countriesToHospitalsField(countriesOfStudy));
```
kduyvesteyn, line 61: I would remove references to the old "countries of study" name

kduyvesteyn (changes requested):
As discussed on call:
 - I would suggest to integrate facility curation into the current `CkbCurator` logic
 - The "CurationFilterEntry" has a confusing name. You explained that while "name" and "zip" curation is sort-of pattern based, the filter entries are for manual curation that do not otherwise fit into the name and zip curation? I am not sure what better names would be in this case, but the term "filter" is at least confusing since you are not actually filtering (removing) stuff. 

Finally, keep in mind due to the changes in ClinicalTrial datamodel this change will be a major upgrade and we have to coordinate the upgrade in ACTIN.

datamodel/src/main/java/com/hartwig/serve/datamodel/Country.java
cbruel: As discussed. Removed cities(), since every city always appears as a key in `hospitalsPerCity()`

algo/src/main/java/com/hartwig/serve/sources/ckb/curation/CkbCurator.java
```diff
@@ -31,45 +50,42 @@ public List<CkbEntry> run(@NotNull List<CkbEntry> ckbEntries) {
...
-    }
-
-    public void reportUnusedCurationEntries() {
-        int unusedEntryCount = 0;
-        for (CurationEntry entry : CurationFactory.VARIANT_MAPPINGS.keySet()) {
-            if (!evaluatedCurationEntries.contains(entry)) {
-                unusedEntryCount++;
-                LOGGER.warn(" Entry '{}' hasn't been used during CKB curation", entry);
+            for (ClinicalTrial clinicalTrial : ckbEntry.clinicalTrials()) {
+                List<Location> locations = Lists.newArrayList();
```
kduyvesteyn, line 58: Maybe rename to `curatedLocations` similar to how `curatedVariants` and `curatedClinicalTrials` are used?

algo/src/main/java/com/hartwig/serve/sources/ckb/curation/CkbCurator.java
```diff
@@ -86,4 +102,85 @@ private Variant curate(@NotNull Variant variant) {
 
         return curatedVariant;
     }
+
+    @NotNull
+    public Location curateFacilityName(@NotNull Location location) {
```
kduyvesteyn, line 107: This function can be package-private (ie drop the `public` keyword). It can't be private because it is used in a test, so would be good to annotate with `VisibleForTesting`

algo/src/main/java/com/hartwig/serve/sources/ckb/curation/CkbCurator.java
```diff
@@ -86,4 +102,85 @@ private Variant curate(@NotNull Variant variant) {
...
+        for (CkbFacilityCurationZipEntry facilityCurationZipEntry : facilityCurationZipEntries) {
+            if (containsWord(facilityCurationZipEntry.city(), location.city().toLowerCase())) {
+                String zip = location.zip() != null ? location.zip().toLowerCase().replaceAll("\\s", "") : "";
+                if ((facilityCurationZipEntry.zip().equals("")) || (zip.contains(facilityCurationZipEntry.zip()))) {
```
kduyvesteyn, line 111: Minor but it is more standard to use `isEmpty` rather than `equals("")`

algo/src/main/java/com/hartwig/serve/sources/ckb/curation/CkbCurator.java
```diff
@@ -86,4 +102,85 @@ private Variant curate(@NotNull Variant variant) {
...
+                        facilityCurationNameEntry.city(),
+                        location.city().toLowerCase())) {
+                    return ImmutableLocation.builder().from(location).facility(facilityCurationNameEntry.curatedFacilityName()).build();
+                }
+            }
+        }
+
+        for (CkbFacilityCurationManualEntry facilityCurationManualEntry : facilityCurationManualEntries) {
+            if (equalStringsOrNull(location.facility(), facilityCurationManualEntry.facilityName()) && location.city()
+                    .equals(facilityCurationManualEntry.city()) && equalStringsOrNull(location.zip(), facilityCurationManualEntry.zip())) {
```
kduyvesteyn, line 129: Another minor but the if statement her is fairly complex. I'd personally split this up in multiple lines, something like:
```
hasMatchingFacilityName = ...
hasMatchingCity = ...
hasMatchingZip = ...
(if hasMatchingFacilityName && hasMatchingCity && hasMatchingZip) {
```

But up to you!

algo/src/main/java/com/hartwig/serve/sources/ckb/curation/CkbCurator.java
```diff
@@ -86,4 +102,85 @@ private Variant curate(@NotNull Variant variant) {
...
+                unusedFacilityCurationManualCount++;
+                LOGGER.warn(" Facility curation manual entry '{}' hasn't been used for CKB filtering", entry);
+            }
+        }
+        LOGGER.debug(" Found {} unused manual facility curation entries during CKB filtering", unusedFacilityCurationManualCount);
+    }
+
+    @VisibleForTesting
+    boolean equalStringsOrNull(@Nullable String string1, @NotNull String string2) {
+        if (string1 == null && string2.equals("")) {
```
kduyvesteyn, line 172: Minor but it is more standard to use `isEmpty` over `equals("")`

algo/src/test/java/com/hartwig/serve/sources/ckb/curation/CkbCuratorTest.java
```diff
@@ -1,33 +1,108 @@
...
     }
+
+    @Test
+    public void canCurateFacilityNameBasedOnName() {
+        Location location1 = createLocation("Rotterdam", "Erasmus MC", "3062 PA");
+        Location location2 = createLocation("Rotterdam", "Erasmus ziekenhuis in Rotterdam", "3062 PA");
+        Location curatedLocation = createLocation("Rotterdam", "EMC", "3062 PA");
+
+        assertEquals(curator.curateFacilityName(location1), curatedLocation);
+        assertEquals(curator.curateFacilityName(location2), curatedLocation);
```
kduyvesteyn, line 47: `assertEquals` expects the expected first and then the actual, so you need to swap around the params!
(Side note: This is one reason why we use `assertThat` in ACTIN, to avoid this confusion!)

Also, I would rename `curatedLocation` to `expectedLocation` since the algo didn't actually curate that location, you create it manually!

algo/src/test/java/com/hartwig/serve/sources/ckb/curation/CkbCuratorTest.java
```diff
@@ -1,33 +1,108 @@
...
+        Location locationWithZip1 = createLocation("Groningen", "Research site", "9713 GZ");
+        Location locationWithZip2 = createLocation("Groningen", "Research site", "9713");
+        Location locationWithoutZip = createLocation("Almere", "Flevo location 2", null);
+        Location curatedLocationWithZip1 = createLocation("Groningen", "UMCG", "9713 GZ");
+        Location curatedLocationWithZip2 = createLocation("Groningen", "UMCG", "9713");
+        Location curatedLocationWithoutZip = createLocation("Almere", "Flevoziekenhuis", null);
+
+        assertEquals(curator.curateFacilityName(locationWithZip1), curatedLocationWithZip1);
+        assertEquals(curator.curateFacilityName(locationWithZip2), curatedLocationWithZip2);
+        assertEquals(curator.curateFacilityName(locationWithoutZip), curatedLocationWithoutZip);
```
kduyvesteyn, line 61: Need to swap around params here.

algo/src/test/java/com/hartwig/serve/sources/ckb/curation/CkbCuratorTest.java
```diff
@@ -1,33 +1,108 @@
...
+    }
+
+    @Test
+    public void canCurateFacilityNameBasedOnFilter() {
```
kduyvesteyn, line 65: Is this test named properly? It seems like no curation happens here? Shouldn't it be something like "skip..."?

algo/src/test/java/com/hartwig/serve/sources/ckb/curation/CkbCuratorTest.java
```diff
@@ -1,33 +1,108 @@
...
+        Location location = createLocation("Amsterdam", "Amsterdam UMC", null);
+
+        assertEquals(curator.curateFacilityName(location), location);
+    }
+
+    @Test
+    public void canFindEqualStringsOrNull() {
```
kduyvesteyn, line 72: Would rename "find" to "evaluate", you are not looking for something :)

algo/src/main/java/com/hartwig/serve/ServeConfig.java
```diff
@@ -89,6 +92,9 @@ static Options createOptions() {
         options.addOption(CKB_BLACKLIST_EVIDENCE_TSV, true, "Path to the CKB blacklist evidence tsv");
         options.addOption(CKB_DRUG_CURATION_TSV, true, "Path to the CKB drug curation tsv");
         options.addOption(CKB_REGION_TSV, true, "Path to the CKB regions tsv. Only trials from the configured regions will be used.");
+        options.addOption(CKB_FACILITY_CURATION_NAME_TSV, true, "Path to the CKB facility curations based on name tsv");
+        options.addOption(CKB_FACILITY_CURATION_ZIP_TSV, true, "Path to the CKB facility curations based on zip tsv");
+        options.addOption(CKB_FACILITY_CURATION_MANUAL_TSV, true, "Path to the CKB facility curations to filter tsv");
```
kduyvesteyn, line 97: This description seems outdated?

algo/src/main/java/com/hartwig/serve/ServeConfig.java
```diff
@@ -250,6 +265,9 @@ static ServeConfig createConfig(@NotNull CommandLine cmd) throws ParseException
                 .ckbBlacklistEvidenceTsv(useCkbEvidence ? nonOptionalFile(cmd, CKB_BLACKLIST_EVIDENCE_TSV) : NOT_APPLICABLE)
                 .ckbDrugCurationTsv(useCkbEvidence ? nonOptionalFile(cmd, CKB_DRUG_CURATION_TSV) : NOT_APPLICABLE)
                 .ckbRegionTsv(useCkbTrials ? nonOptionalFile(cmd, CKB_REGION_TSV) : NOT_APPLICABLE)
+                .ckbFacilityCurationNameTsv(useCkbTrials ? nonOptionalFile(cmd, CKB_FACILITY_CURATION_NAME_TSV) : NOT_APPLICABLE)
+                .ckbFacilityCurationZipTsv(useCkbTrials ? nonOptionalFile(cmd, CKB_FACILITY_CURATION_ZIP_TSV) : NOT_APPLICABLE)
+                .ckbFacilityCurationManualTsv(useCkbTrials ? nonOptionalFile(cmd, CKB_FACILITY_CURATION_MANUAL_TSV) : NOT_APPLICABLE)
```
kduyvesteyn, line 270: This is tricky because in ServeAlgo you use these params also if only `useCkbEvidence` is true. 

I think I would check for both (`useCkbTrials || useCkbEvidence`) in this config

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -123,16 +127,18 @@ && hasPotentiallyOpenRequirementToInclude(trial.recruitment()) && hasAgeGroupToI
...
-        }
-
-        return countries;
+    static Set<Country> extractCountriesToInclude(@NotNull ClinicalTrial trial, @NotNull Set<CkbRegion> regionsToInclude) {
+        Map<String, Map<String, Set<String>>> countriesToCitiesToHospitalNames = trial.locations()
+                .stream()
+                .filter(location -> regionsToInclude.stream().anyMatch(region -> region.includes(location))
+                        && hasPotentiallyOpenRequirementToInclude(location.status()))
+                .collect(Collectors.groupingBy(Location::country,
+                        Collectors.groupingBy(Location::city, Collectors.mapping(Location::facility, Collectors.toSet()))));
```
kduyvesteyn, line 136: I suspect this may even impress John :)

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -184,53 +184,108 @@ public void shouldNotCreateAnActionableEntryWhenVariantRequirementIsOnADifferent
...
+        assertEquals(0, ActionableTrialFactory.extractCountriesToInclude(trialSuspendedInBelgiumOpenInIndia, REGIONS_TO_INCLUDE).size());
+
+        ClinicalTrial terminatedTrial = createTrialWithOneLocation("Terminated", "Netherlands", "Rotterdam", "Suspended", null);
+        assertEquals(0, ActionableTrialFactory.extractCountriesToInclude(terminatedTrial, REGIONS_TO_INCLUDE).size());
+    }
 
-        ClinicalTrial terminatedTrial = createTrialWithOneLocation("Terminated", "Netherlands", "Suspended");
-        assertEquals(0, ActionableTrialFactory.filterOnRegionsToInclude(terminatedTrial, REGIONS_TO_INCLUDE).size());
+    @Test
+    public void canAddCities() {
```
kduyvesteyn, line 265: Would it be cleaner to name something like "canMapCities"?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableTrialFactoryTest.java
```diff
@@ -184,53 +184,108 @@ public void shouldNotCreateAnActionableEntryWhenVariantRequirementIsOnADifferent
...
+        ClinicalTrial trialWithMultipleLocations = createTrialWithMultipleLocations("Recruiting",
+                "Belgium",
+                "Recruiting",
+                "Brussel",
+                "UZ Brussel",
+                "Netherlands",
+                "Not yet recruiting",
+                "Rotterdam",
+                "EMC");
+        assertEquals(ActionableTrialFactory.extractCountriesToInclude(trialWithMultipleLocations, REGIONS_TO_INCLUDE),
```
kduyvesteyn, line 283: Params need to be swapped (first expected, then actual)

datamodel/src/test/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtilTest.java
kduyvesteyn, line 1: You test all the individual functions but it might be nice to add one overarching function. 

Or more concretely I would propose:
 - Rename `canConvertActionableEvents` to `canConvertActionableTreatment`
 - Add a similar function `canConvertActionableTrial` and create a test trial that is "complicated" (multiple cities with multiple hospitals per city) and check if you can convert it back-and-forth

kduyvesteyn (approved):
Changes look all good!! Have done more in-depth review compared to last time, added a bunch of thoughts/suggestions, and also made a few changes myself along the way (see recent commits)

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
cbruel: haha I will be honest, Mr Chat Geepeetee helped me a bit here

PR #38, opened 2024-08-28
ACTIN-1217: Add new fields to serve to support expanded evidence model

kzuberihmf:


kzuberihmf:
Test output dataset using these changes in my home dir on actin-serve vm.

kzuberihmf:
So consuming the output this change in an version of ACTIN with corresponding updates seemed to work. First time attempting a serve update, so likely doing some things wrong 😄.

Updated ActionableEvent adding `date` and `description` attributes, corresponding to the update date and efficacy evidence from ckb. I made these nullable which is maybe against conventions thought I'd try it out. For the tsv these are written/read from blank fields. 

If we want non-nullable, the date field would be a bit ugly with some placeholder value, but description as an non-nullable empty string would work I suppose.

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtil.java
```diff
@@ -189,6 +211,8 @@ public static String toLine(@NotNull ActionableEvent event) {
                 .add(event.level().toString())
                 .add(event.direction().toString())
                 .add(setToField(event.evidenceUrls()))
+                .add(formatDate(event.date()))
```
cbruel, line 214: Curious why you make a function `formatDate` here, but for the line below you don't (although it's the same "kind of" code). I would think the `formatDate` function is not needed either, and you could just inline the code?

cbruel (commented):
Looks good to me. Only 1 small comment. 

What are you planning to use the "description" field for? Those are very long to show on the ACTIN report..

cbruel:
Two more comments (based on discussion with Korneel):
- You added the 2 new fields at the bottom of `ActionableEvent`, would it be more logical to change the order? E.g. add `date` below `evidenceUrls`
- As you already mentioned, it would be better to make the fields @NotNull (they are also @NotNull in ckb datamodel). And then for the other sources (vicc, iclusion) use an empty string or some placeholder value. We do that already in e.g. `iclusion/ActionableTrialFactory` where we set level always to EvidenceLevel.B and direction always to EvidenceDirection.RESPONSIVE.

kzuberihmf:
> Two more comments (based on discussion with Korneel):
> 
> * You added the 2 new fields at the bottom of `ActionableEvent`, would it be more logical to change the order? E.g. add `date` below `evidenceUrls`

Yes I put the two fields at the bottom because optimal order wasn't obvious to me 😄. But I may be misunderstanding your recommendation, because in `ActionableEvent` the date() field is already below evidenceUrls()?

> * As you already mentioned, it would be better to make the fields @NotNull (they are also @NotNull in ckb datamodel). And then for the other sources (vicc, iclusion) use an empty string or some placeholder value. We do that already in e.g. `iclusion/ActionableTrialFactory` where we set level always to EvidenceLevel.B and direction always to EvidenceDirection.RESPONSIVE.

Will try not null, thanks.

cbruel:
> > * You added the 2 new fields at the bottom of `ActionableEvent`, would it be more logical to change the order? E.g. add `date` below `evidenceUrls`
> 
> Yes I put the two fields at the bottom because optimal order wasn't obvious to me 😄. But I may be misunderstanding your recommendation, because in `ActionableEvent` the date() field is already below evidenceUrls()?
> 

Oeps my fault, I meant above, but I also didn't think it through. I think this order would make the most sense: source(), date(), sourceEvent(), sourceUrls(), intervention(), applicableCancerTypes(), blacklistCancerTypes(), description(), level(), direction(), evidenceUrls()

kzuberihmf:
Trying the `EPOCH` as the date placeholder, which is sort of the zero-value date & corresponds to Jan 1 1970. Already a built-in standard constant, and should stand out in usage if it shows up in an unintended place.

Will try generating beta output on serve and consuming in ACTIN, usually find I messed up something 😆.

kduyvesteyn (approved):
Haven't reviewed in detail, the changes you made based on Charlotte feedback sound all good to me! Also, it seems the build is failing so there may be something wrong somewhere still :)

kzuberihmf:
> Haven't reviewed in detail, the changes you made based on Charlotte feedback sound all good to me! Also, it seems the build is failing so there may be something wrong somewhere still :)

Thanks for having a look. The build is failing for the docker image, something to do with python2 vs python3. The maven artifact builds and is uploaded correctly however (and so is usable by ACTIN). Will ask around about the docker setup, maybe someone recognizes the issue as it seems vaguely familiar.

kzuberihmf:
Docker build resolved, need to stick to a base image version that supports python2. Thanks to @nedleitch for reminding me of the solution.

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/util/ActionableFileUtil.java
```diff
@@ -174,6 +196,7 @@ public static String toLine(@NotNull ActionableEvent event) {
             therapy.add(treatment.name());
         }
         return new StringJoiner(FIELD_DELIMITER).add(event.source().toString())
+                .add(event.date().toString())
```
cbruel, line 199: I think we would like to keep the field empty in the tsv if the date is LocalDate.EPOCH?

datamodel/src/test/resources/actionable/ActionableCharacteristics.37.tsv
```diff
@@ -1,3 +1,3 @@
...
-MICROSATELLITE_UNSTABLE			CKB_TRIAL	msi	http	NCT1	study1		both	Netherlands(Rotterdam;Eindhoven),Belgium(Brussel)	Rotterdam(EMC),Eindhoven(Catharina),Brussel(UZ Brussel)	Pembro,Immuno			All cancer types	X	Hematologic cancer;2531	B	GUIDELINE	RESPONSIVE	http3,http4
\ No newline at end of file
+type	cutoffType	cutoff	source	date	sourceEvent	sourceUrls	nctId	title	acronym	genderCriterium	countriesAndCities	hospitalsPerCity	treatment	treatmentApproachesDrugClass	treatmentApproachesTherapy	applicableCancerType	applicableDoid	blacklistCancerTypes	description	level	approvalStatus	direction	evidenceUrls
+MICROSATELLITE_UNSTABLE	EQUAL_OR_GREATER	4	CKB_EVIDENCE	2021-02-03	msi	http							Nivolumab			All cancer types	X	Hematologic cancer;2531	efficacy evidence	A	GUIDELINE	RESPONSIVE	http1,http2
+MICROSATELLITE_UNSTABLE			CKB_TRIAL	1970-01-01	msi	http	NCT1	study1		both	Netherlands(Rotterdam;Eindhoven),Belgium(Brussel)	Rotterdam(EMC),Eindhoven(Catharina),Brussel(UZ Brussel)	Pembro,Immuno			All cancer types	X	Hematologic cancer;2531		B	GUIDELINE	RESPONSIVE	http3,http4
```
cbruel, line 3: See comment above. Don't think it's preferred to have 1970-01-01 in the output files. We could set this to an empty string?

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -154,6 +154,8 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
                             .approvalStatus(approvalStatus)
                             .direction(direction)
                             .evidenceUrls(evidenceUrls)
+                            .date(entry.updateDate())
```
cbruel, line 157: Would be nice to keep the same order here as in the datamodel? source, date etc.
Same in some other places you altered (don't know if this is really a coding standard, but I think it's nice :) )

datamodel/src/test/resources/actionable/ActionableCharacteristics.37.tsv
kzuberihmf: Ok, though this looks suspiciously like handling a nullable field 😄. Anyway I'll update, but how should the db dao handle this? I've made the field non-null sql DATETIME field there and it would be populated with the EPOCH value. Should this be nullable in mysql? Or prefer empty strings to DATETIME fields there?

kduyvesteyn: I think I would keep in the 1970 date rather than empty. Don't think we should do specific serialisation / deserialisation logic. As for the SQL model, I think I would write 1970 as well since we should do as-little-as-possible conversion (the SQL data should simply be a view on the data)

cbruel: Ok, let's just leave it this way :)

cbruel (approved):

PR #41, opened 2024-09-20
ACTIN-1294: Added efficacy evidence date to SERVE, extract most recent of all available dates + refactoring of evidence properties

spdeleeuw:
- Renamed date to ckbEntryDate to keep it clear that it's different from the efficacyDescriptionDate field. 
- Renamed description to efficacyDescription
- Year related to evidence is allways the efficacyDescriptionYear -> i.e. the year of the corresponding scientific publication. If that field is null, we take the most recent of therapyCreateDate and ckbEntryDate

Bonus: Korneels datamodel feedback on ACTIN-1248


spdeleeuw:
@kzuberihmf you are absolutely right! I just started adjusting that :)

kzuberihmf (commented):
Regarding the new names, are we ok with having a ckb specific column name? I thought the idea was to be a bit more generic so as to apply to other sources, but I may be mistaken.

pauldwolfe (commented):
Looks fine, but given @kduyvesteyn and @kzuberihmf on this PR will leave it to them to approve

spdeleeuw:
@kzuberihmf @kduyvesteyn @pauldwolfe about the last commits:
- Processed Khalid's feedback
- I found out that apart from `null` the year field of evidenceReference can also be `0`. Now that's incorporated in the year-extraction function.

datamodel/src/main/java/com/hartwig/serve/datamodel/ActionableEvent.java
```diff
@@ -29,7 +29,10 @@ public interface ActionableEvent {
     Set<CancerType> blacklistCancerTypes();
 
     @NotNull
-    String description();
+    String efficacyDescription();
+
+    @NotNull
+    String efficacyDescriptionYear();
```
kduyvesteyn, line 35: This looks odd as a String. Shouldn't it a nullable integer?

datamodel/src/test/java/com/hartwig/serve/datamodel/DatamodelTestFactory.java
```diff
@@ -99,27 +99,30 @@ public static ActionableEvent createTestActionableEvent() {
...
                 EvidenceLevel.A,
                 EvidenceLevelDetails.FDA_APPROVED,
                 EvidenceDirection.NO_BENEFIT,
-                Sets.newHashSet()
-        );
+                Sets.newHashSet());
     }
 
     @NotNull
     public static ActionableEvent createActionableEvent(@NotNull Knowledgebase source, @NotNull LocalDate date, @NotNull String sourceEvent,
```
kduyvesteyn, line 110: Would rename `date` to `entryDate` here to match with datamodel

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -259,6 +258,28 @@ static EvidenceDirection resolveDirection(@Nullable String direction) {
         return null;
     }
 
+    @NotNull
+    @VisibleForTesting
+    static String extractEvidenceYear(@NotNull LocalDate entryDate, @NotNull List<Reference> references, Therapy therapy) {
```
kduyvesteyn, line 263: I am quite sure you can make this into a @Nullable Integer!

Also, therapy should be annotated as `Nullable`

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
```diff
@@ -259,6 +258,28 @@ static EvidenceDirection resolveDirection(@Nullable String direction) {
...
+                .filter(year -> year != null && !year.isEmpty() && !year.equals("0"))
+                .max(Comparator.naturalOrder());
+
+        if (mostRecentYear.isPresent()) {
+            return mostRecentYear.get();
+        }
+
+        int year;
+        if (therapy != null) {
+            year = !entryDate.isBefore(therapy.createDate()) ? entryDate.getYear() : therapy.createDate().getYear();
```
kduyvesteyn, line 275: I would reverse the evaluation here (`entryDate.isBefore(therapy.createDate) ?`). 

Also, you could inline this (`int year = `) though that may make reading the code a bit harder, so no strong opinion.

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableTrialFactory.java
```diff
@@ -98,8 +98,9 @@ public Set<ActionableEntry> create(@NotNull CkbEntry entry, @NotNull String sour
                                     .evidenceLevelDetails(EvidenceLevelDetails.CLINICAL_STUDY)
                                     .direction(EvidenceDirection.RESPONSIVE)
                                     .evidenceUrls(Sets.newHashSet("https://clinicaltrials.gov/study/" + trial.nctId()))
-                                    .date(LocalDate.EPOCH)
-                                    .description(Strings.EMPTY)
+                                    .entryDate(LocalDate.EPOCH)
+                                    .efficacyDescription(Strings.EMPTY)
+                                    .efficacyDescriptionYear(Strings.EMPTY)
```
kduyvesteyn, line 103: Some thoughts here:
 - Would suggest to set the fields in the order of the datamodel
 - Why not use the actual entry date here?
 - If you always set the efficacy fields to Empty, maybe they belong in `Treatment` rather than `ActionableEvent`?

algo/src/main/java/com/hartwig/serve/sources/iclusion/ActionableTrialFactory.java
```diff
@@ -85,8 +85,9 @@ private static ImmutableActionableTrial.Builder toActionableBuilder(@NotNull Icl
                 .evidenceLevelDetails(EvidenceLevelDetails.UNKNOWN)
                 .direction(EvidenceDirection.RESPONSIVE)
                 .evidenceUrls(Sets.newHashSet())
-                .date(LocalDate.EPOCH)
-                .description(Strings.EMPTY);
+                .entryDate(LocalDate.EPOCH)
+                .efficacyDescription(Strings.EMPTY)
+                .efficacyDescriptionYear(Strings.EMPTY);
```
kduyvesteyn, line 90: Set the fields in the order of the datamodel

algo/src/main/java/com/hartwig/serve/dao/ServeDAO.java
```diff
@@ -141,7 +141,8 @@ private void writeActionableHotspots(@NotNull Timestamp timestamp, @NotNull List
                     ACTIONABLEHOTSPOT.APPLICABLECANCERTYPE,
                     ACTIONABLEHOTSPOT.APPLICABLEDOID,
                     ACTIONABLEHOTSPOT.BLACKLISTCANCERTYPES,
-                    ACTIONABLEHOTSPOT.DESCRIPTION,
+                    ACTIONABLEHOTSPOT.EFFICACYDESCRIPTION,
+                    ACTIONABLEHOTSPOT.EFFICACYDESCRIPTIONYEAR,
                     ACTIONABLEHOTSPOT.LEVEL,
                     ACTIONABLEHOTSPOT.EVIDENCELEVELDETAILS,
```
kduyvesteyn, line 147: Unrelated to this ticket, but it's weird that the evidence level is called "level" and the evidence level details are called "evidence level details". This is not consistent (but should have been caught in an earlier ticket!)

algo/src/main/java/com/hartwig/serve/dao/ServeDAO.java
```diff
@@ -218,7 +219,7 @@ private static void writeActionableHotspotBatch(@NotNull Timestamp timestamp, @N
                 actionableHotspot.applicableCancerType().name(),
                 actionableHotspot.applicableCancerType().doid(),
                 concat(toStrings(actionableHotspot.blacklistCancerTypes())),
-                actionableHotspot.description(),
+                actionableHotspot.efficacyDescription(),
```
kduyvesteyn, line 222: You don't seem to write the year?

This seems to be cause in a number of functions in this class?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
kduyvesteyn, line 1: Should the general tests also test for the year?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
```diff
@@ -293,4 +314,38 @@ private static CkbEvidenceBlacklistModel createBlacklistModel(@NotNull CkbBlackl
...
+                .createDate(LocalDate.of(createYear, 1, 1))
+                .updateDate(LocalDate.of(2021, 1, 1))
+                .therapyName("Therapy")
+                .drugs(List.of())
+                .synonyms(List.of())
+                .globalApprovalStatuses(List.of())
+                .description("Description")
+                .references(List.of())
+                .build();
+    }
```
kduyvesteyn, line 351: Did you consider moving these functions into `CkbTestFactory`? 

This factory currently mainly creates entries, so may not be a good fit but at least something to consider

datamodel/src/test/java/com/hartwig/serve/datamodel/serialization/ActionableGeneFileTest.java
kduyvesteyn, line 1: Would suggest to assert the fields in the order of the datamodel

datamodel/src/test/java/com/hartwig/serve/datamodel/serialization/ActionableHotspotFileTest.java
kduyvesteyn, line 1: Assert the fields in the order of the datamodel

datamodel/src/test/java/com/hartwig/serve/datamodel/serialization/ActionableRangeFileTest.java
kduyvesteyn, line 1: Assert the fields in the order of the datamodel

kduyvesteyn (commented):
General thoughts:
 - I doubt the SQL works since your inserter has the year but you don't write it. Did you test this?
 - The jira ticket suggests that you write the therapy entry date, but instead you write the (profile) entry date. What do you intend to do? I personally think the profile entry date is better, but in that case please update the jira ticket!

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
spdeleeuw: I think the return value of the function should remain NotNull right? Since you fall back on entryDate if there is no referenceDate and no therapy(Date)?

datamodel/src/main/java/com/hartwig/serve/datamodel/ActionableEvent.java
spdeleeuw: I agree on the Integer part, but shouldn't it remain @NotNull since we fall back on therapy and entry dates for this value?

kduyvesteyn: If it is always there it should be notnull indeed. Wasn't really sure when I wrote this comment.

One minor note, the type should just be `int` rather than `notnull integer`, since that's effectively what it is (a primitive type cannot be null so should be used instead of objects in case of notnull)

algo/src/main/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactory.java
kduyvesteyn: My suggestion wasn't complete I think :)

Note that `isAfter` is not the same as `!isBefore` when the dates are equal (but in this use-case it doesn't matter. 

A cleaner implementation could be: `Math.max(therapy.createDate.year(), entryDate.getYear)` in case therapy != null?

algo/src/test/java/com/hartwig/serve/sources/ckb/ActionableEvidenceFactoryTest.java
```diff
@@ -184,6 +194,21 @@ public void canResolveDirections() {
...
+        List<Reference> referencesWithDate = List.of(CkbTestFactory.createReference("2021"), CkbTestFactory.createReference("2023"));
+        List<Reference> referencesWithoutDate = List.of(CkbTestFactory.createReference(null));
+        Therapy therapy = CkbTestFactory.createTherapy(2018);
+        LocalDate mostRecentDate = LocalDate.of(2024, 1, 1);
+
+        assertEquals(Integer.valueOf(2023), ActionableEvidenceFactory.extractEvidenceYear(entryDate, referencesWithDate, therapy));
+        assertEquals(Integer.valueOf(2023), ActionableEvidenceFactory.extractEvidenceYear(mostRecentDate, referencesWithDate, therapy));
+        assertEquals(Integer.valueOf(2018), ActionableEvidenceFactory.extractEvidenceYear(entryDate, referencesWithoutDate, therapy));
+        assertEquals(Integer.valueOf(2024), ActionableEvidenceFactory.extractEvidenceYear(mostRecentDate, referencesWithoutDate, therapy));
+        assertEquals(Integer.valueOf(2015), ActionableEvidenceFactory.extractEvidenceYear(entryDate, referencesWithoutDate, null));
```
kduyvesteyn, line 209: When you change to `int` you can also remove the `Integer.valueOf()` here

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbTestFactory.java
```diff
@@ -206,4 +208,38 @@ private static Gene createGene(@NotNull String geneSymbol) {
                 .geneRole("")
                 .build();
     }
+
+    @NotNull
+    public static Reference createReference(String year) {
```
kduyvesteyn, line 213: `year` is missing annotation (notnull/nullable)

kduyvesteyn (approved):
Changes look good to me!

Do have one new thought: is "efficacy description year"  the best name for this property? It may suggest "the year that the efficacy description was written" which is not correct (or at least not generic enough). Maybe a name like `evidenceYear` would be cleaner? Or `efficacyYear`?

PR #42, opened 2024-10-04
ACTIN-1320: Json serialization for SERVE

kzuberihmf:
Are there other SERVE users that need updating for the json change? Currently this PR add's JSON serialization without removing the TSV's just in case.

* Introduces ServeRecord, containing known and actionable events, and ServeJson with serialization/deserialization helpers
* Produces two separate json files: serve.37.json and serve.38.json

Creating a full SERVE release results in json files that are each about 65M. 

The previous extraction result output produces a couple of vcf's and a file `EventInterpretation.tsv` that was not a part of the serve_db tables, not sure where used but these also remain, and EventInterpretation is not part of the json.


datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/ServeJson.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import com.hartwig.serve.datamodel.RefGenome;
+import com.hartwig.serve.datamodel.ServeRecord;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class ServeJson {
+    private static final ObjectMapper mapper = new ObjectMapper()
+            .registerModule(new GuavaModule())
```
kzuberihmf, line 19: Because somehow google collections are used in to the datamodel in de-serialization from json

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/ServeJson.java
```diff
@@ -0,0 +1,45 @@
...
+            .registerModule(new JavaTimeModule())
```
kzuberihmf, line 20: Date's are serialized to json like this:

```
      "entryDate": [2016,1,19],
```

Which is the serializers default behaviour

algo/src/main/java/com/hartwig/serve/extraction/ExtractionResultWriter.java
```diff
@@ -43,6 +48,11 @@ public ExtractionResultWriter(@NotNull final String outputDir, @NotNull final Re
     }
 
     public void write(@NotNull ExtractionResult result) throws IOException {
+        writeTsv(result);
```
pauldwolfe, line 51: suggestion, we could also do this with a feature toggle approach where you turn on JSON support? What do you think @cbruel @kduyvesteyn ?

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/ServeJson.java
pauldwolfe: I think you can also add `@Value.Style(jdkOnly = true)` to force immutables to not use guava.

algo/src/main/java/com/hartwig/serve/extraction/ExtractionResultWriter.java
pauldwolfe: The advantage being you wouldn't need to cart around TSV if you are using JSON and vice-versa

kduyvesteyn: The TSV should be removed, shouldn't be a need to have them anymore.

We do need to keep the hotspot VCF and the interpretation TSV btw, but do think they could be separate outputs (they are not needed by the typical SERVE users

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/ServeJson.java
```diff
@@ -0,0 +1,45 @@
...
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.datatype.guava.GuavaModule;
+import com.fasterxml.jackson.datatype.jsr310.JavaTimeModule;
+import com.hartwig.serve.datamodel.RefGenome;
+import com.hartwig.serve.datamodel.ServeRecord;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class ServeJson {
+    private static final ObjectMapper mapper = new ObjectMapper()
```
kduyvesteyn, line 18: Should be `MAPPER` rather than `mapper`

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/ServeJson.java
kduyvesteyn: Why is this different from ACTIN itself where we use a lot of dates and also JSON serialization/deserialization? 
E.g. could we reuse things from `GsonSerializer` from ACTIN repo?

kduyvesteyn (approved):
One general question: Does the SERVE algo application produce a JSON that is guaranteed to be the same upon 2 runs of SERVE? This used to be quite annoying and the reason why there are so many quite detailed comparators in the datamodel package. 

It is a convenient situation to be able to diff 2 jsons from different versions and same input and not have to remove sorting differences.

kduyvesteyn:
Some notes on what files are produced by SERVE:
 - Known events and actionable events (per ref genome version)
 - Event interpretation which is purely for verification / debugging (SERVE has a 2-step approach -> first every profile from a source KB is "interpreted" and gets assigned a type, and then based on that type further information is extracted. However, the "profile to type mapping" isn't part of the output datamodel and doesn't have to be, but for testing it can be convenient to have this mapping for a source KB. 
 - known hotspots VCF -> These are the exact same hotspots but generated in VCF format as SAGE expects this input to be in VCF format. SAGE could be refactored to take a SERVE output but doubt that is a good idea (it only needs the known hotspots anyways and VCF is quite an established format

cbruel (commented):
I mainly focused on checking the output in your `/home/kzuberi/ACTIN-1320` folder on `serve-vm` (comparing the tsv and json). Looking at the files in `out_nohotspot`: “countries” is always empty, which is (obviously) not correct.. Could you look into why this field is not written correctly to the json?
Rest looks good to me, thanks for making this change!

kzuberihmf:
> I mainly focused on checking the output in your `/home/kzuberi/ACTIN-1320` folder on `serve-vm` (comparing the tsv and json). Looking at the files in `out_nohotspot`: “countries” is always empty, which is (obviously) not correct.. Could you look into why this field is not written correctly to the json? Rest looks good to me, thanks for making this change!

Thanks for checking the data! That output was actually generated from a slightly older version of the update that didn't have the country annotation. I had produced a full version of the output that did check out. However to be sure I've just updated the no_hotspot version, and double checked that the countries are populated. So should be good.

kzuberihmf:
Updates:

* Removed guava and added jdkOnly annotations, thanks for the suggestion @pauldwolfe 
* Revised date format as above comment
* Added general sorting during serialization. Applies to Sets, Maps, and Lists.  Updated the datamodel objects to make them comparable directly for this to work. 
  * note resorting lists is slightly worrying, if we ever have a list we don't want reordered
  * implementation is more fiddly than I'd hoped, found some trick online
  * wonder if we could instead we depend on jq to re-sort when doing manual comparisons? 
* deleted tsv files writing, but maintained vcf's and event interpretation output file

kduyvesteyn (approved):
Haven't reviewed in depth, the comments on what has changed all sound good to me.

It may be a nice simplification to directly pass `serve_json` into LoadServeDatabase rather than 2 params which are then internally resolved to the serve json, but fairly minor.

cbruel (approved):
Thanks for changing the output data! Looks good. I really like the json output, much cleaner than the tsv :) (especially for the countries).

datamodel/src/main/java/com/hartwig/serve/datamodel/serialization/ServeJson.java
pauldwolfe: We're moving away from GSON in general and prefer Jackson. The rationale is:
- In Java projects, Jackson has out of the box integration with @Immutables.
- Jackson provides more extensions and flexibility without having to define adapters. 
- Jackson is the library of choice for all web frameworks (jax-rs, resteasy, etc)

PR #43, opened 2024-10-05
ACTIN-1325: Make country and city nullable in CKB datamodel and mapping

kduyvesteyn:
@cbruel Could you please review if facility curation in `CkbCurator` still works as expected?


algo/src/test/java/com/hartwig/serve/sources/ckb/curation/CkbCuratorTest.java
```diff
@@ -47,6 +47,12 @@ public void canCurateFacilityNameBasedOnName() {
         assertEquals(expectedLocation, curator.curateFacilityName(location2));
     }
 
+    @Test
+    public void canCurateFacilityWithMissingCity() {
+        Location missingCity = createLocation(null, "facility", null);
+        assertEquals(missingCity, curator.curateFacilityName(missingCity));
+    }
```
kzuberihmf, line 54: I suppose country could also be null, if you want to include that in the test.

kzuberihmf (approved):
LGTM

algo/src/test/java/com/hartwig/serve/sources/ckb/curation/CkbCuratorTest.java
kduyvesteyn: good one, added

cbruel:
@kduyvesteyn Looks good!

PR #44, opened 2024-10-09
ACTIN-1336: Add version to SERVE json output

kzuberihmf:
I'm not actually sure what the downstream intended usage is, as I don't recall seeing us include app versions in outputs other than log files. But easy enough to add assuming it makes sense, here's a small update.

Branched off of my earlier branch, hence the long commit history.  Also cleaned up some left over test tsv files.

Update: Motivation is part of the general goal of being able to trace all outputs from the pipeline to the software versions that created them, which will now be possible for SERVE.


kduyvesteyn (approved):
Looks good!

LiekeSchoenmaker (approved):
Looks good, the adding serve version

PR #45, opened 2024-10-18
ACTIN-1205: Redesign SERVE datamodel to split trials from evidence and support complex profiles

cbruel:
Fully tested


datamodel/src/main/java/com/hartwig/serve/datamodel/ClinicalTrial.java
kduyvesteyn, line 1: Unrelated but I wonder where `genderCriterium` could be an enum? It is hard to interpret downstream as a string.

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeData.java
```diff
@@ -0,0 +1,27 @@
...
+public abstract class ServeData {
+
+    @NotNull
+    abstract KnownEvents knownEvents();
+
+    @NotNull
+    abstract List<ActionableTreatment> actionableTreatments(); // I'm doubting this is correct
+
+    @NotNull
+    abstract List<ClinicalTrial> clinicalTrials();
```
kduyvesteyn, line 26: Functions should all be public

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeData.java
kduyvesteyn, line 1: May sound odd but I think I'd rename this to `ServeRecord` :) 

The SERVE database then contains one record per ref genome. The term "data" feels a bit too abstract.

datamodel/src/main/java/com/hartwig/serve/datamodel/MolecularEvent.java
kduyvesteyn, line 1: This may sound contrary to discussion on friday but maybe `ActionableEvent` is a good name for this still. It represents the base entity of a molecular event that is actually actionable. Now that treatment and trial don't extend this anymore I feel this may actually be a nice name?

kduyvesteyn (commented):
See comments! 

I feel the general model is quite good already, and my feeling is that your trial model fits with CKB's well. 

You could actually try to make datamodel package by itself compilable, that would make it easier to review the datamodel as a whole (`mvn clean package -pl datamodel -am`)

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeDatabase.java
```diff
@@ -0,0 +1,24 @@
...
+
+@Value.Immutable
+@Value.Style(passAnnotations = { NotNull.class, Nullable.class },
+             jdkOnly = true)
+@JsonSerialize(as = ImmutableServeDatabase.class)
+@JsonDeserialize(as = ImmutableServeDatabase.class)
+public abstract class ServeDatabase {
+
+    @NotNull
+    public abstract String serveVersion();
```
kduyvesteyn, line 20: Would rename to just `version`, the "serve" part is implicit

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeDatabase.java
```diff
@@ -0,0 +1,24 @@
...
+
+    @NotNull
+    public abstract Map<RefGenome, ServeRecord> serveDataPerRefGenome();
```
kduyvesteyn, line 23: Would rename to `records`, the "serve" part is implicit. "recordsPerRefGenome" could also work but I think that is also considered implicit, and I like concise names at the top of the datamodel but that's personal

datamodel/src/main/java/com/hartwig/serve/datamodel/EfficacyEvidence.java
```diff
@@ -0,0 +1,47 @@
...
+    public abstract int evidenceYear();
+
+    @NotNull
+    public abstract Set<String> evidenceUrls();
+
+    @NotNull
+    public abstract Indication indication();
+
+    @NotNull
+    public abstract MolecularProfile molecularProfile();
```
kduyvesteyn, line 46: I would put these variables more up top, I think I'd order treatment -> indication -> profile -> efficacyDescription -> ...

datamodel/src/main/java/com/hartwig/serve/datamodel/ClinicalTrial.java
```diff
@@ -32,5 +32,14 @@ public abstract class ClinicalTrial implements Intervention {
...
+    public abstract GenderCriterium genderCriterium();
+
+    @NotNull
+    public abstract Set<String> urls();
+
+    @NotNull
+    public abstract Set<Indication> indications();
+
+    @NotNull
+    public abstract Set<MolecularProfile> molecularProfiles();
```
kduyvesteyn, line 44: I would put these more up top, though not all the way. I would put at least `urls` at the bottom.

Another suggestion, though not super strong opinion, would `molecularCriteria` be a better term for the profiles? It matches nicely with `genderCriterium` and is more specific overall

datamodel/src/main/java/com/hartwig/serve/datamodel/EfficacyEvidence.java
```diff
@@ -0,0 +1,47 @@
...
+    @NotNull
+    public abstract EvidenceLevelDetails evidenceLevelDetails();
+
+    @NotNull
+    public abstract EvidenceDirection evidenceDirection();
+
+    public abstract int evidenceYear();
+
+    @NotNull
+    public abstract Set<String> evidenceUrls();
```
kduyvesteyn, line 40: Maybe rename to just `urls` similar to how the urls are called in clinical trial?

This variable used to have `efficacy` in there I believe because there were also other types of urls in the datamodel before (if I recall correctly)

datamodel/src/main/java/com/hartwig/serve/datamodel/ActionableEventComparator.java
```diff
@@ -8,65 +8,16 @@ public class ActionableEventComparator implements Comparator<ActionableEvent> {
...
+        if (dateCompare != 0) {
+            return dateCompare;
         }
 
-        int directionCompare = event1.direction().toString().compareTo(event2.direction().toString());
-        if (directionCompare != 0) {
-            return directionCompare;
+        int urlsCompare = event1.sourceUrls().toString().compareTo(event2.sourceUrls().toString());
+        if (urlsCompare != 0) {
+            return urlsCompare;
```
kduyvesteyn, line 18: Minor, but would sorting on event and then urls be more logical?

datamodel/src/main/java/com/hartwig/serve/datamodel/EfficacyEvidenceComparator.java
kduyvesteyn, line 1: I am not sure this comparison is exhaustive. E.g. if you have identical efficacy evidence for 2 different molecular profiles, this comparator would consider them equal?

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeDatabase.java
```diff
@@ -0,0 +1,24 @@
...
+             jdkOnly = true)
+@JsonSerialize(as = ImmutableServeDatabase.class)
+@JsonDeserialize(as = ImmutableServeDatabase.class)
+public abstract class ServeDatabase {
+
+    @NotNull
+    public abstract String version();
+
+    @NotNull
+    public abstract Map<RefGenome, ServeRecord> records();
```
kzuberihmf, line 23: Just pointing out I this means we'll parse all the json only to usually (or always?) ignore half of it, and the ServeRecord is a bit fat at approaching 70M. But I can see how its convenient from a data management point of view, so ok.

datamodel/src/main/java/com/hartwig/serve/datamodel/ClinicalTrial.java
```diff
@@ -32,5 +32,14 @@ public abstract class ClinicalTrial implements Intervention {
...
-    public abstract String genderCriterium();
-}
\ No newline at end of file
+    public abstract GenderCriterium genderCriterium();
+
+    @NotNull
+    public abstract Set<Indication> indications();
+
+    @NotNull
+    public abstract Set<MolecularCriterium> molecularCriteria();
```
kzuberihmf, line 41: Dumb question, the semantics are these must all be satisfied, or any? I recall seeing some ticket mentioning both complex and combined profiles as different things and wasn't sure how they needed to be modelled.

kzuberihmf (commented):
Looks good to me in general, thanks for taking this on!

Found the ticket I mentioned, a bit old but in case it's still useful in thinking about the model: https://hartwigmedical.atlassian.net/browse/OSP-42

datamodel/src/main/java/com/hartwig/serve/datamodel/ClinicalTrial.java
cbruel: For a clinical trial you should read this set as 'OR', since for example for a trial you should have an EGFR amp OR an EGFR fusion). So not all `Set<MolecularCriterium>` have to be satisfied.

If we have combined evidence (e.g. “ALK fusion” +  “ALK G1202R”) this will be in 1 `MolecularCriterium` with in `fusions()` one `ActionableFusion` and in `hotspots()` one `ActionableHotspot`. So _within_ a `MolecularCriterium` you should read it as 'AND'.

Didn't know about the ticket, thanks for pointing me to it. 
Complex evidence as explained in this ticket, is not modelled by this new datamodel. Me talking about "complex profiles" before was probably confusing, let's talk about "combined evidence" from now on when we talk about e.g. “ALK fusion” +  “ALK G1202R” :)

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeDatabase.java
cbruel: Thanks for bringing this up, I didn't think about this technical stuff (and so also have no opinion on it). Will leave it like this for now, if you say it's ok

kduyvesteyn: @kzuberihmf In practice we may get the situation where we get tests with different ref genomes, e.g. if we switch internally to 38 while the external tests done by other labs are on 37 then you need to load both SERVE versions. I don't think we want to add this complexity to outside of the application.

If the size is a problem we could make smart loaders potentially? E.g. if ACTIN-molecular starts it could first take the unique set of all ref genome versions of molecular inputs and then only load the ones it needs? Could also optimize within `ServeRecord` (e.g. trial ingestion only needs the known events). Though sounds like something for later if it turns out to be an actual problem?

datamodel/src/main/java/com/hartwig/serve/datamodel/ClinicalTrial.java
kduyvesteyn: In SERVE datamodel we use "combined evidence" while CKB uses "complex profiles" for the same thing, so that may cause the confusion :) 

Wondering if we could clarify the fact that the criteria should be or'ed in the name, don't have inspiration though (`anyMolecularCriteria`?)

datamodel/src/main/java/com/hartwig/serve/datamodel/ServeDatabase.java
kzuberihmf: Agree to leave until it becomes an actual problem.

kduyvesteyn (approved):
A piece of art!

PR #48, opened 2024-11-11
ACTIN-1457 Extract "likelihood type" from driver panel as gene role

pauldwolfe:
Also add the same field to the curated genes.

Open questions:
- Should we rename likekihoodType in the gene panel to gene role?
- Should we continue to use the same extractor for gene panel and curated genes?
- Should we add gene role to cohort and curated hotspots?


kduyvesteyn (approved):
Looks good though I would split up HartwigGeneExtractor into a HartwigCuratedGeneExtractor and a HartwigDriverGeneExtractor since for the curated  input we will call the field "geneRole" while for driver gene extractor we have to stick with "likelihoodType"

kduyvesteyn:
> Also add the same field to the curated genes.
> 
> Open questions:
> 
> * Should we rename likekihoodType in the gene panel to gene role?
> * Should we continue to use the same extractor for gene panel and curated genes?
> * Should we add gene role to cohort and curated hotspots?

1: Would not rename in the driver panel. From driver perspective likelihood type is a better name. We assume it is identical to gene role but that's not a claim the driver panel makes.
2: Suggest to split them up (see also my comment on the approval itself)
3: It would be clean to do so yet added value minimal as I am quite sure every cohort and curated hotspot is on a gene in our driver panel, so the information would flow in through that route anyway.

PR #49, opened 2024-11-13
ACTIN-1466: Include gene role in known genes comparator

kduyvesteyn:


datamodel/src/main/java/com/hartwig/serve/datamodel/gene/KnownGeneComparator.java
```diff
@@ -8,6 +8,11 @@ public class KnownGeneComparator implements Comparator<KnownGene> {
 
     @Override
     public int compare(@NotNull KnownGene gene1, @NotNull KnownGene gene2) {
-        return gene1.gene().compareTo(gene2.gene());
+        int geneCompare = gene1.gene().compareTo(gene2.gene());
```
pauldwolfe, line 11: Java SDK offers some nice chaining of comparators:

```suggestion
        return Comparator.comparing(KnownGene::gene).thenComparing(KnownGene::geneRole).compare(gene1,
```

pauldwolfe (approved):
Looks good, one improvement suggestion.

datamodel/src/main/java/com/hartwig/serve/datamodel/gene/KnownGeneComparator.java
kduyvesteyn: Tempting but will leave as-is, there are a lot of comparators in SERVE and would rather refactor them in one go instead of having inconsistent ways

PR #50, opened 2024-11-13
ACTIN-1467: Split up driver gene and curated gene inputs

kduyvesteyn:
Made a few changes in addition to changes from ACTIN-1457:
 - Remove exposing a field name of a specific input file in ServeAlgo.
 - Enforce GeneRole datamodel when reading input file.


algo/src/main/java/com/hartwig/serve/sources/hartwig/gene/HartwigCuratedGeneFileReader.java
```diff
@@ -0,0 +1,19 @@
+package com.hartwig.serve.sources.hartwig.gene;
+
+import java.io.IOException;
+import java.util.List;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class HartwigCuratedGeneFileReader {
```
pauldwolfe, line 8: You could also create a named static factory method to reduce the duplication between HartwigCuratedGeneFileReader and HartwigDriverGeneFileReader and keep the explicit naming.

pauldwolfe (approved):

algo/src/main/java/com/hartwig/serve/sources/hartwig/gene/HartwigCuratedGeneFileReader.java
kduyvesteyn: Good one, updated

PR #51, opened 2024-11-28
ACTIN-1525: Remove google.common dependency from datamodel

jbartletthmf:


datamodel/src/test/java/com/hartwig/serve/datamodel/molecular/characteristic/TumorCharacteristicComparatorTest.java
```diff
@@ -22,7 +22,9 @@ public void canSortTumorCharacteristics() {
                 create(TumorCharacteristicType.MICROSATELLITE_UNSTABLE, TumorCharacteristicCutoffType.GREATER, 4D);
         TumorCharacteristic characteristic4 = create(TumorCharacteristicType.MICROSATELLITE_UNSTABLE, null, null);
 
-        List<TumorCharacteristic> characteristics = Lists.newArrayList(characteristic1, characteristic3, characteristic4, characteristic2);
+        List<TumorCharacteristic> characteristics = new ArrayList<>(
+            List.of(characteristic1, characteristic3, characteristic4, characteristic2)
+        );
```
kduyvesteyn, line 27: You can drop the new ArrayList here I think since you already use list.of

datamodel/src/test/java/com/hartwig/serve/datamodel/molecular/common/ChromosomeComparatorTest.java
```diff
@@ -19,7 +18,7 @@ public void canSortChromosomes() {
         String chromosome5 = "Y";
         String chromosome6 = "chrM";
 
-        List<String> chromosomes = Lists.newArrayList(chromosome5, chromosome4, chromosome1, chromosome6, chromosome2, chromosome3);
+        List<String> chromosomes = new ArrayList<>(List.of(chromosome5, chromosome4, chromosome1, chromosome6, chromosome2, chromosome3));
```
kduyvesteyn, line 21: Drop the new ArrayList here

kduyvesteyn (commented):
There are lots of instances of `new ArrayList<>(List.of(...))` which seems wrong to me? Have commented on 2 of them but there are >10.

datamodel/src/test/java/com/hartwig/serve/datamodel/molecular/characteristic/TumorCharacteristicComparatorTest.java
jbartletthmf: Most of these changes support comparator tests, which require a mutable list to be populated with known values. `List.of` produces an immutable list and the `ArrayList` constructor only accepts a collection at initialization time, so this was the best option I came up with. Would you prefer something else? E.g.
1. `list = new ArrayList<>(); list.add(c1); list.add(c2);`
2. `list = new ArrayList<>(); List.of(c1, c2).forEach(list::add);`
3. `list = Stream.of(c1, c2).collect(Collectors.toList());` (The IDE actually prompts you to sort the stream before collection in this case, which makes a lot of sense, but I'm not sure we want to change the precondition of an unsorted list.)

jbartletthmf: Another possibility would be to use the immutable list as input and create a new one after sorting:
```suggestion
        List<TumorCharacteristic> characteristics = List.of(characteristic1, characteristic3, characteristic4, characteristic2);
        List<TumorCharacteristic> sorted = characteristics.stream().sorted(new TumorCharacteristicComparator()).collect(Collectors.toList());
```
This is similar to the stream-with-sorting approach above, but creates an initial list that can be considered separately. This approach makes sense if we want to take a small step towards using more immutable collections in SERVE.

kduyvesteyn: Ah I didn't realize this. I thought I tried to simplify on one place and didn't see an error elsewhere so assumed it was possible to simply remove. 

I don't have a strong preference on the approach.

kduyvesteyn: yes go for it!

PR #52, opened 2024-12-09
ACTIN-1281: Assign EvidenceLevel to decreased response evidence

cbruel:
(See email thread _"Re: "no benefit" vs "not predictive" evidence -> can we interpret these identically?"_ from September 10th)

CKB does not assign a level/tier in case:
- response type: conflicting, decreased response, not predictive, unknown
- evidence type: emerging, risk factor
- approval status: preclinical - biochemical

All evidence without evidence level is filtered in SERVE. This is acceptable/what we want for above, except for ‘decreased response’ (since we want to be able to show decreased response evidence on the ACTIN report). Therefore we should assign an evidence level to ‘decreased response' based on EvidenceLevelDetails:
- PRECLINICAL, CASE_REPORTS_SERIES → D
- CLINICAL_STUDY → C
- FDA_APPROVED, GUIDELINE, FDA_CONTRAINDICATED → A

### DIFF
--- a/algo/README.md
+++ b/algo/README.md
@@ -293,6 +293,9 @@ is mapping evidence for tumor characteristics (such as MSI or High TMB) to actua
 SERVE only considers trials with one or more molecular inclusion criterion. Furthermore, SERVE only considers trials with a potentially open
 recruitment status ("Recruiting", "Not yet recruiting", "approved for marketing", "available" or "enrolling by invitation") and required
 requirement type ("partial - required" and "required").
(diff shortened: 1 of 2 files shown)


kduyvesteyn (approved):
Looks good!

In general may be an idea to have more proper documentation for some of the interpretations we do on CKB side. E.g. in this case we infer a value that CKB itself doesn't dare to infer directly, and it would be good to capture the rationale behind this somewhere (other than "see email...")

PR #53, opened 2024-12-17
ACTIN-1570: Patch SQL datamodel to v7

kduyvesteyn:
The key changes:
 - There are now `efficacyEvidence` and `actionableTrial` table
 - the `trialMolecularCriterium` table maps every trial to N molecular criteria. 
 - the model supports combined evidence.

Views may be helpful to make the data more visible without complex queries but is for later.

I have tested loading 7.2.0 ACTIN NL output and will test loading 7.2.0 public sources output.

NOTE: I have removed some performance optimization (batching) since I am not sure if it works with chained SQL commands. The loading now takes ~10 min rather than ~1 min which I think is fine in practice anyways.


algo/src/main/java/com/hartwig/serve/dao/EventInterpretationDAO.java
```diff
@@ -0,0 +1,49 @@
...
+        }
+    }
+
+    private static void writeEventInterpretationBatch(@NotNull InsertValuesStep5 inserter,
+            @NotNull EventInterpretation eventInterpretation) {
+        inserter.values(eventInterpretation.source().name(),
+                eventInterpretation.sourceEvent(),
+                eventInterpretation.interpretedGene(),
+                eventInterpretation.interpretedEvent(),
+                eventInterpretation.interpretedEventType());
```
kzuberihmf, line 47: interpretedEventType().name() ?

algo/src/main/java/com/hartwig/serve/dao/ServeDAO.java
```diff
@@ -1,1218 +1,69 @@
...
-        for (Indication indication : baseTrial.indications()) {
-            expandedTrials.add(trialBuilder.indications(List.of(indication)).build());
-        }
-        return expandedTrials;
-    }
+    public void repopulate(@NotNull ServeRecord serveRecord, @NotNull List<EventInterpretation> eventInterpretations) {
+        LOGGER.info("Deleting all data from SERVE database");
+        knownEventsDAO.deleteAll();
+        eventInterpretationDAO.deleteAll();
+        molecularCriteriumDAO.deleteAll();
```
kzuberihmf, line 53: I don't see molecularCriteriumDAO.write() anywhere below

PR #57, opened 2025-02-25
ACTIN-1737: Group HOTSPOTS in SERVE datamodel

kzuberihmf:
No functional changes, just grouping.

Looked consistent matching hotspots counts between this and the current master serve version.


algo/src/main/java/com/hartwig/serve/dao/MolecularCriteriumDAO.java
```diff
@@ -46,7 +46,8 @@ public int write(@NotNull MolecularCriterium molecularCriterium) {
         int molecularCriteriumId =
                 context.insertInto(MOLECULARCRITERIUM).values().returning(MOLECULARCRITERIUM.ID).fetchOne().getValue(MOLECULARCRITERIUM.ID);
 
-        writeActionableHotspots(molecularCriteriumId, molecularCriterium.hotspots());
+        // TODO do we need a subset id to model the hotspot grouping?
```
kzuberihmf, line 49: Currently can't reverse this to recover the groups, is that needed?

If so, not sure we want to introduce a further auto-generated db id, could generate an index count in the insertion code, if that doesn't seem to weird.

datamodel/src/main/java/com/hartwig/serve/datamodel/molecular/hotspot/ActionableHotspotSet.java
kduyvesteyn, line 1: Would it be simpler to make something like 
`public class ActionableHotspotSet implements Set<ActionableHotspot>` 

If I am not mistaken this saves one level of depth when you need to get to the individual hotspots and still removes the "set of set" construction somewhat?

EDIT: Not confident this would make it better, but somehow the "hotspots.hotspots" feels odd to me as well, so must be able to come up with something better!

datamodel/src/main/java/com/hartwig/serve/datamodel/molecular/MolecularCriterium.java
```diff
@@ -26,7 +26,7 @@ public abstract class MolecularCriterium implements Comparable<MolecularCriteriu
     private static final Comparator<MolecularCriterium> COMPARATOR = new MolecularCriteriumComparator();
 
     @NotNull
-    public abstract Set<ActionableHotspot> hotspots();
+    public abstract Set<ActionableHotspotSet> hotspots();
```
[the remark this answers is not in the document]
kduyvesteyn: See my comment in ActionableHotspotSet.

algo/src/main/java/com/hartwig/serve/dao/MolecularCriteriumDAO.java
kduyvesteyn: I'd think this is fine for now.

Might be nice to make a ticket for further improvement though, some thoughts:
 - We dont really need an ID generated in the database for molecular criterium, nor do we need the table. We could also just use an index count in the insertion code.
 - We should try to multi-thread the insertion of the data. The SERVE insertion used to be really fast but due to my recent changes this is no longer the case. I couldn't figure out how to multi-thread the insertion of the molecular criteria.

kduyvesteyn (commented):
I don't think this is ideal yet, see my comments in PR. 

Some additional thoughts:
 - Maybe we could have ActionableHotspot not extend VariantHotspot anymore but just contain a set of variants, and it could still extend ActionableEvent. This data for every variant must be identical so ugly to duplicate.
 - I haven't tested this extensively but my feeling is that EventExtractorResult could maybe stop returning Lists everywhere. I think in practice it will always return max 1 codon and max 1 exon, and only a list of Variants, but this is basically always the set of variants that makes up a single ActionableHotspot. 

All in all I feel we can do a better job with representing "one actionable hotspot can be represented as a set of mutually exclusive variants for which one needs to be present"

datamodel/src/main/java/com/hartwig/serve/datamodel/molecular/hotspot/ActionableHotspotSet.java
kzuberihmf: Played around with this, but had some trouble getting it to interact correctly with immutables + serialization.

kzuberihmf:
Follow-up on correctness, 

> I don't think this is ideal yet, see my comments in PR.
> 
> Some additional thoughts:
> 
> * Maybe we could have ActionableHotspot not extend VariantHotspot anymore but just contain a set of variants, and it could still extend ActionableEvent. This data for every variant must be identical so ugly to duplicate.

Seems like a way forward, will wait for @pauldwolfe to chime in before implementing.

> * I haven't tested this extensively but my feeling is that EventExtractorResult could maybe stop returning Lists everywhere. I think in practice it will always return max 1 codon and max 1 exon, and only a list of Variants, but this is basically always the set of variants that makes up a single ActionableHotspot.
> 

👍 

> All in all I feel we can do a better job with representing "one actionable hotspot can be represented as a set of mutually exclusive variants for which one needs to be present"

kzuberihmf:
Minor note re previous concerns on mismatch of the count of evidences in the prototype. Since this version changes the model but holds back the combined evidences its easier to compare. Now while the  count of evidences don't match, when adding up all the hotspots in the hotspot groups the counts match up exactly with the current SERVE version. So I think that worry is resolved.

kzuberihmf:
Thanks for doing this! Seems nice to me, will do some regression.

kzuberihmf:
I ran some regression on the kd branch, including with VICC, and all seemed consistent. Will double-check tomorrow but I expect that its ok.

Is the idea to merge this work to master, make a SERVE release and update matching logic in ACTIN, or complete the series of subsequent updates for the combined profiles before making a release?

kduyvesteyn:
I'd suggest to make a release after this datamodel change and support the new datamodel in ACTIN. 

This could be done in parallel.

kzuberihmf:
Regression with the `-kd`  on ckb with nl config, V37:

- known hotspots match exactly 142229
- efficacy evidences`-kd`: 24527, `master`:  47864. But when expanding the the count to include the sum of grouped variants, `-kd` produces 47864 so matches exactly
- actionable hotspots in sql table when loading from `-kd`: 35011, from `master`: 35011

switching to VICC, by loading using the `run_serve_actin_public_sources_only` script
- known hotspots match exactly  9445
- efficacy evidences `-kd`: 12003, `master`: 37962, counting variants for `-kd`: 37962 so matches
- didn't reload the sql in this case, I think we should be good

kzuberihmf:
> @kzuberihmf @pauldwolfe Have given the idea of ActionableHotspot not extend VariableHotspot anymore a try in a new branch ACTIN-1737-kd and quite like it! If you like it as well feel free to merge into original branch.

Merged, thanks!

PR #58, opened 2025-02-28
Actin 1737 kd

pauldwolfe:


datamodel/pom.xml
```diff
@@ -41,6 +41,10 @@
             <artifactId>junit</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.jooq</groupId>
+            <artifactId>jooq</artifactId>
+        </dependency>
```
kzuberihmf, line 47: doesn't seem needed afaict

PR #60, opened 2025-03-20
ACTIN-1768: Implement combined molecular criteria for SERVE efficacy evidence

kzuberihmf:


kzuberihmf:
Still a draft because I'm trying to track down a strange bug. Doing regression tests I see some evidences that I think should have combined molecular criteria but have only a simple criteria. I couldn't find the error in the extraction logic, but finally think I've found it: After the extraction is done there is a conversion to V37 from V38, and does some filtering that can drop molecular criteria. 

I think previously since there would only be a simple criteria this would result in the evidence being dropped. However now with combined criteria, if one criteria is filtered but other parts of the combination remain then the evidence remains but with incorrect criteria since part is missing!

Does that sound right @kduyvesteyn @cbruel? A clue is that a case I found dropped involves a fusion in NCOA4, and that gene appears in `refgenome.ConversionFilterFactory.GENES_TO_EXCLUDE_FOR_CONVERSION`.  So I think that conversion logic still needs to be reviewed, with some rule saying roughly something like: if the criteria count has changed after conversion but isn't 0, then drop the evidence.

kzuberihmf:
I made a narrow change to the evidence conversion filter. Now a combined event should remain in the source genome where it is valid, but be completely dropped from the target genome when one part of the combination is dropped.  Will have to update for trials etc later.

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
```diff
@@ -49,55 +55,83 @@ public void canExtractEvidenceAndTrialsFromCkbEntries() {
...
+        assertEquals(2, extractionResult.knownEvents().genes().size());
+        assertEquals(2, extractionResult.knownEvents().copyNumbers().size());
+        assertEquals(0, extractionResult.knownEvents().fusions().size());
+
+        assertEquals(1, extractionResult.eventInterpretations().size());
+        EventInterpretation interpretation = extractionResult.eventInterpretations().iterator().next();
+        assertEquals(EventType.COMBINED, interpretation.interpretedEventType());
+        assertEquals("Multiple", interpretation.interpretedGene());
+        assertEquals("loss,loss", interpretation.interpretedEvent());
+        assertEquals("Multiple loss,loss", interpretation.sourceEvent());
```
kduyvesteyn, line 97: This is fine but just wanted to note this is not super informative :)

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
```diff
@@ -49,55 +55,83 @@ public void canExtractEvidenceAndTrialsFromCkbEntries() {
...
 
-        List<CodonAnnotation> curatedCodons = CkbExtractor.curateCodons(extractorOutput).codons();
+    @Test
+    public void shouldReturnNullForUnresolvableEvents() {
+        CkbExtractor ckbExtractor = CkbExtractorFactory.createExtractor(CkbClassificationConfig.build(),
+                RefGenomeResourceTestFactory.buildTestResource37(),
+                TreatmentApproachTestFactory.createEmptyCurator(),
+                CkbFilteringTestFactory.createEmptyEvidenceFilterModel(),
+                CkbFilteringTestFactory.createEmptyTrialFilterModel(),
+                Set.of(ImmutableCkbRegion.builder().country("netherlands").states(Collections.emptySet()).build()));
```
kduyvesteyn, line 111: Duplicate code?

datamodel/src/main/java/com/hartwig/serve/datamodel/util/MolecularCriteriumCombiner.java
```diff
@@ -0,0 +1,38 @@
...
+import com.hartwig.serve.datamodel.molecular.ImmutableMolecularCriterium;
+import com.hartwig.serve.datamodel.molecular.MolecularCriterium;
+
+import org.jetbrains.annotations.NotNull;
+
+public final class MolecularCriteriumCombiner {
+    @NotNull
+    public static MolecularCriterium combine(@NotNull List<MolecularCriterium> criteria) {
+        return criteria.stream().reduce(MolecularCriteriumCombiner::combine)
+                .orElse(ImmutableMolecularCriterium.builder().build());
```
kduyvesteyn: Why do we need `orElse`? In case the list of criteria is empty? It feels odd to return an empty criterium in that case, can't we check that upstream and/or return null?

datamodel/src/main/java/com/hartwig/serve/datamodel/util/MolecularCriteriumCombiner.java
kduyvesteyn, line 1: Why did you decide to place this file in datamodel package rather than algo package?

kzuberihmf:
> All my suggestions are minor and/or refactoring.
> 
> In terms of impact/testing, is it the following correct?
> 
> * The actionable trials are unchanged
> * The event interpretation is largely unchanged (only changes to interpreted fields for combined events)
> * There are several 1000 new evidences (less than the total number of combined profile, but majority does get ingested) and evidences that were previously there are still there.

Some counts of things, comparing this branch vs master after applying to filtering fix mentioned above, on last weeks data, V37.

Trials: 402 in both this version and master. Also comparing contents and not just counts, the trial contents are exactly identical.

Evidences: master: 24745, this branch: 38869. In more detail:
* 24450 evidences are exactly the same in both
* this branch adds 14419 evidences, all have combined criteria
* this brach loses 295 single-criteria evidences that are in master. Investigating one by hand, it was the case of an incorrect interpretation of a self-fusion where master was getting 8 single-criteria evidences, which were removed and replaced by a single combined criteria in this branch. Both are wrong (ACTIN-1807). I haven't examined all of these by hand but can investigate more. I vaguely recall there may be e.g. a KIT mis-interpretation also but have to check.

For the Interpretations, master and this branch share 50752 identical interpretations. However something looks suspicious with the new version, new combined events may not be being added correctly. I've actually looked at this the least, probably because its not directly used in ACTIN. Will investigate.

Thanks for the other feedback, will clean up a couple things!

kduyvesteyn:
> > All my suggestions are minor and/or refactoring.
> > In terms of impact/testing, is it the following correct?
> > 
> > * The actionable trials are unchanged
> > * The event interpretation is largely unchanged (only changes to interpreted fields for combined events)
> > * There are several 1000 new evidences (less than the total number of combined profile, but majority does get ingested) and evidences that were previously there are still there.
> 
> Some counts of things, comparing this branch vs master after applying to filtering fix mentioned above, on last weeks data, V37.
> 
> Trials: 402 in both this version and master. Also comparing contents and not just counts, the trial contents are exactly identical.
> 
> Evidences: master: 24745, this branch: 38869. In more detail:
> 
> * 24450 evidences are exactly the same in both
> * this branch adds 14419 evidences, all have combined criteria
> * this brach loses 295 single-criteria evidences that are in master. Investigating one by hand, it was the case of an incorrect interpretation of a self-fusion where master was getting 8 single-criteria evidences, which were removed and replaced by a single combined criteria in this branch. Both are wrong (ACTIN-1807). I haven't examined all of these by hand but can investigate more. I vaguely recall there may be e.g. a KIT mis-interpretation also but have to check.
> 
> For the Interpretations, master and this branch share 50752 identical interpretations. However something looks suspicious with the new version, new combined events may not be being added correctly. I've actually looked at this the least, probably because its not directly used in ACTIN. Will investigate.
> 
> Thanks for the other feedback, will clean up a couple things!

This sounds acceptable to me in terms of verification. We should review at some point whether any event could generate a list of codon or exon annotations, I am not sure about that. 

What do you mean with "new combined events may not be being added correctly"?

kzuberihmf:
> > What do you mean with "new combined events may not be being added correctly"?
> 
> I think I meant to say I that the EventInterpretation.tsv produced by this branch seemed to have too few entries. I was a pretty fried last night when I wrote this.

Further follow-up on the EventInterpretation. This branch does have fewer entries, but I think this is correct.

In the master version, all events are placed in the interpretation tsv, including those that are not processed for various reasons, such as being combined, or complex. In this new version, I only put the events that we can process into the interpretation tsv, hence there are actually fewer entries now.

Is this desired, or do we want all the rejected events included in the interpretation tsv again? These would be combined events where part of the combination failed interpretation, or complex events that we do not support.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractor.java
```diff
@@ -0,0 +1,200 @@
...
+import com.hartwig.serve.extraction.fusion.FusionConsolidation;
+import com.hartwig.serve.extraction.variant.KnownHotspotConsolidation;
+
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+final class CkbKnownEventsExtractor {
+
+    @NotNull
+    public static KnownEvents generateKnownEvents(@NotNull List<ExtractedEvent> allEvents, boolean efficacyEvidencesIsEmpty) {
```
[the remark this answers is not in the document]
kzuberihmf: I moved this here for maintainability, hopefully not duplicating 🤔

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbExtractorTest.java
kzuberihmf: I think I just tried to keep this consistent with the original interpretation. Can change as preferred.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -120,291 +69,167 @@ private ExtractionResult getExtractionResult(@NotNull CkbEntry entry) {
...
+        EventInterpretation interpretation = interpretEvent(entry, extractedEvents);
 
-    private void addFusionsToCriteria(@NotNull EventExtractorOutput extractionOutput, @NotNull ActionableEvent actionableEvent,
-            @NotNull Set<MolecularCriterium> molecularCriteria) {
-        if (extractionOutput.fusionPair() != null) {
-            Set<ActionableFusion> fusions = extractActionableFusions(extractionOutput.fusionPair(), actionableEvent);
-            molecularCriteria.add(ImmutableMolecularCriterium.builder().fusions(fusions).build());
-        }
-    }
+        Set<MolecularCriterium> molecularCriteria = Set.of(CkbMolecularCriteriaExtractor.createMolecularCriterium(entry, extractedEvents));
```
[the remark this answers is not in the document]
kzuberihmf: Did this, yes was cleaner. Will likely change further in the next round when trials are updated for combined evidence.

datamodel/src/main/java/com/hartwig/serve/datamodel/util/MolecularCriteriumCombiner.java
kzuberihmf: Hmm, maybe not justified, but thought having an api to combine might be useful to users of the package (though the only use I can think of right now is test cases)

datamodel/src/main/java/com/hartwig/serve/datamodel/util/MolecularCriteriumCombiner.java
kzuberihmf: I guess I didn't like spelling empty in two different ways (empty criterium and null), might have done that in another place as well. Might be worth changing but think I'll leave it right now.

algo/src/main/java/com/hartwig/serve/sources/ckb/filter/CkbMolecularProfileFilterModel.java
```diff
@@ -32,24 +34,19 @@ public CkbMolecularProfileFilterModel(@NotNull final List<CkbMolecularProfileFil
...
-                } else {
-                    LOGGER.debug("Filtering variant '{}' on '{}'", variant.variant(), variant.gene().geneSymbol());
-                }
+                Map<Boolean, List<Variant>> shouldIncludeVariants = entry.variants().stream()
+                        .collect(Collectors.partitioningBy(variant -> include(entry.type(), variant)));
 
-                if (!filteredVariants.isEmpty()) {
-                    cleanedCkbEntries.add(ImmutableCkbEntry.builder().from(entry).variants(filteredVariants).build());
+                if (shouldIncludeVariants.get(false).isEmpty()) {
+                    cleanedCkbEntries.add(ImmutableCkbEntry.builder().from(entry).variants(shouldIncludeVariants.get(true)).build());
```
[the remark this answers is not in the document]
kzuberihmf: I think I'm not going to touch this logic right now actually.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -120,291 +69,167 @@ private ExtractionResult getExtractionResult(@NotNull CkbEntry entry) {
...
+                    .interpretedGene(gene)
+                    .interpretedEvent(event)
+                    .interpretedEventType(entry.type())
+                    .build();
+        } else {
+            return ImmutableEventInterpretation.builder()
+                    .source(Knowledgebase.CKB)
+                    .sourceEvent("Multiple " + concat(entry.variants()))
+                    .interpretedGene("Multiple")
+                    .interpretedEvent(concat(entry.variants()))
```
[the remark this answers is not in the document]
kzuberihmf: Yeah this is a good point. Since I'm now always extracting interpretations, I've made it so that extractedEvents will always be the full list so the count will be consistent. As for the choice of using the original entry here, I think this was due to just trying very hard to be consistent with the previous (which might have actually been arbitrary).

I can change this, but will be easier to do one last quick regression with everything as consistent as possible first.

kzuberihmf:
I think this is just about ready. One quick test run to make sure nothing was damaged and the the interpretations are complete now.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractor.java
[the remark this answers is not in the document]
kduyvesteyn: I meant why do we check for `efficacyEvidencesIsEmpty` just for known codons, that feels arbitrary

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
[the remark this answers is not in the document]
kduyvesteyn: Clear. It does feel weird/wrong to have `sourceEvent` and `interpretedEvent` not be a simple concat of multiple of the same entries. IBut if making that change makes regression harder we can leave for later

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
```diff
@@ -120,291 +68,180 @@ private ExtractionResult getExtractionResult(@NotNull CkbEntry entry) {
...
-    private void addExonsToCriteria(@NotNull EventExtractorOutput extractionOutput, @NotNull ActionableEvent actionableEvent,
-            @NotNull Set<MolecularCriterium> molecularCriteria) {
-        if (extractionOutput.exons() != null) {
-            Set<ActionableRange> exons = extractActionableRanges(extractionOutput.exons(), actionableEvent);
-            for (ActionableRange exon : exons) {
-                molecularCriteria.add(ImmutableMolecularCriterium.builder().exons(Set.of(exon)).build());
-            }
+        List<ExtractedEvent> emptyEventOutputs = extractedEvents.stream()
+                .filter(e -> countEventExtractorOutputs(e.eventExtractorOutput()) == 0)
+                .collect(Collectors.toList());
```
kduyvesteyn, line 97: This can be moved higher since it is also tested in the first if statement?

kduyvesteyn (approved):
Just a few minor comments remaining, looks good!

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbExtractor.java
kzuberihmf: > I meant why do we check for efficacyEvidencesIsEmpty just for known codons, that feels arbitrary

Yeah I don't know about that, previous behavior, seems to have been introduced in ACTIN-953. Can follow up with the author, but leaving for now.

PR #61, opened 2025-04-01
ACTIN-1840: Remove all variants from CKB with UNKNOWN impact when we map them to known events

cbruel:
Have started a serve run to check results


algo/src/test/java/com/hartwig/serve/sources/ckb/CkbTestFactory.java
kduyvesteyn, line 1: proteinEffect is nullable in the CKB datamodel, might be clean to support that in the test factory as well? 

Also, that way you don't have to pass empty string for the MSI-high evidence in `CkbExtractorTest` line 177

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractor.java
kduyvesteyn, line 1: I would be in favour of making this change inside `convertToKnownSet` so that it impacts codons/exons/copynumbers/fusions as well - that logic should be general (we want no known fusion with an unknown protein effect from CKB into SERVE output).

Also, there seems to be no unit-test at all for this class, that is worrying! Not sure where that slipped our attention or whether this is tested somewhere else but would be good to add tests in case this logic is not tested .

cbruel: It's tested (indirectly) in `CkbExtractorTest`. But I now added a separate test class

cbruel:
Thanks for reviewing! Will kick off a new serve run to test

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractorTest.java
```diff
@@ -0,0 +1,131 @@
...
+        extractedEvents.add(createCodon("BRAF", "V600", "gain of function"));
+        extractedEvents.add(createCodon("TP53", "R175", "unknown"));
+        extractedEvents.add(createExon("BRAF", "exon 1 deletion", "loss of function"));
+        extractedEvents.add(createExon("APC", "exon 3 deletion", "unknown"));
+        extractedEvents.add(createFusion("ALK", "EML4", "gain of function"));
+        extractedEvents.add(createFusion("APC", "BRAF", "unknown"));
+        extractedEvents.add(createHotspot("BRAF", "V600E", "gain of function"));
+        extractedEvents.add(createHotspot("TP53", "R175H", "unknown"));
+        extractedEvents.add(createAmplification("MET", "gain of function"));
+        extractedEvents.add(createAmplification("KIT", "unknown"));
```
kduyvesteyn, line 39: Minor but my brain likes the ordering of the test to be the same as the ordering in the logic. In SERVE we typically evaluate things "from small to big", so hotspots -> codons -> exons -> genes -> fusions.

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractorTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @NotNull
+    private static ExtractedEvent createExon(@NotNull String gene, @NotNull String event, @Nullable String proteinEffect) {
+        return ImmutableExtractedEvent.builder()
+                .gene(gene)
+                .event(event)
+                .variant(CkbTestFactory.createVariant(gene, event, gene + " " + event, proteinEffect))
+                .eventType(EventType.EXON)
+                .eventExtractorOutput(ImmutableEventExtractorOutput.builder()
+                        .addExons(ImmutableExonAnnotation.builder()
```
kduyvesteyn, line 84: I would use `RangeTestFactory.knownExonBuilder` here, that way you don't have to specify input transcript and input exon rank, since they don't matter for your test

algo/src/test/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractorTest.java
```diff
@@ -0,0 +1,131 @@
...
+
+    @NotNull
+    private static ExtractedEvent createAmplification(@NotNull String gene, @Nullable String proteinEffect) {
+        return ImmutableExtractedEvent.builder()
+                .gene(gene)
+                .event(gene + " amp")
+                .variant(CkbTestFactory.createVariant(gene, "amp", gene + " amp", proteinEffect))
+                .eventType(EventType.AMPLIFICATION)
+                .eventExtractorOutput(ImmutableEventExtractorOutput.builder()
+                        .copyNumber(ImmutableGeneAnnotationImpl.builder().gene(gene).event(GeneEvent.AMPLIFICATION).build())
```
kduyvesteyn, line 127: I would use `GeneTestFactory.createGeneAnnotation` here. Even though the impact is minor it is good to minimise the creation of immutables to the least amount of places, that makes it relatively easier to add a field.

algo/src/main/java/com/hartwig/serve/sources/ckb/CkbKnownEventsExtractor.java
```diff
@@ -196,4 +188,25 @@ private static Set<KnownFusion> convertToKnownFusions(@Nullable FusionPair fusio
                 CkbVariantAnnotator::annotateFusion,
                 variant);
     }
+
+    @NotNull
+    private static <T, U> Set<U> convertToKnownSet(@Nullable List<T> rawList, @NotNull Function<T, U> convert,
```
kduyvesteyn, line 178: Per advice, it would make this code more readable if you define that `U` has to extend `KnownEvent`. That way it makes things a bit more readable on what `e` could be in your filter function.

You do this by starting with `private static <T, U extends KnownEvent> convertToKnownSet(...`

Note that this has no actual code impact, just makes it clearer for reader what `U` could be.

kduyvesteyn (approved):
Looks good, just some code comments!

PR #62, opened 2025-04-04
ACTIN-1855: Broaden sourceEvent column

cbruel:
Due to the new combined evidence, we get an error loading the data into sql because the data doesn't fit into the sourceEvent column anymore (tested in pilot and is solved)


kduyvesteyn (approved):
Looks good! Note that I want to take a look at the formatting of the source events, think we can make it a bit nicer, but can be considered independent of this.

PR #63, opened 2025-05-20
ACTIN-1999: Actually apply the converted actionable hotspots when performing ref genome conversion

kduyvesteyn:
Since `8.0.0` we no longer effectively convert actionable hotspots from one ref genome to another which is a pretty serious bug. 

I think this PR should fix it - currently running an exhaustive SERVE run on serve-vm-actin-1 VM, will be finished and checked tomorrow.

### DIFF
--- a/algo/README.md
+++ b/algo/README.md
@@ -447,6 +447,8 @@ elsewhere.
 
 ## Version History and Download Links
 
+- Upcoming
+  - Fix in SERVE ref genome conversion for actionable hotspots to actually apply the lifted-over hotspots in the conversion.
 - [8.3.0](https://github.com/hartwigmedical/serve/releases/tag/serve-v8.3.0)
     - Remove all variants from CKB with UNKNOWN protein impact when we map them to known events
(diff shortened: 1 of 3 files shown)


ArnoHMF (commented):
Looks good to me, but I'm not familiar enough with the code to understand why this would fix the issue. 

I'm assuming that it's fixed because you make the copy of the hotspot?

ArnoHMF:
Ah clear, lifted was indeed never set on the return value. 

The test is a bit subtle. I would expect something that really tests that you are using x instead of chrx. But I suppose this works as well.

kduyvesteyn:
> Ah clear, lifted was indeed never set on the return value.
> 
> The test is a bit subtle. I would expect something that really tests that you are using x instead of chrx. But I suppose this works as well.

Agree the test is not ideal, but the logic that turns `x` into `chrX` is external and requires quite some resource files to be created. At least now the test checks that the output object != input object, better than before

kzuberihmf (approved):
😞 Thanks for catching!

PR #70, opened 2026-02-24
ACTIN-3199: Add ageGroups to evidence datamodels and filter on adult/senior evidence

cbruel:
CKB added new "ageGroups" field to their evidence datamodel

### DIFF
--- a/algo/README.md
+++ b/algo/README.md
@@ -449,7 +449,7 @@ elsewhere.
 ## Version History and Download Links
 
 - [Upcoming]
-  - Add ageGroups to evidence datamodels
+  - Add ageGroups to evidence datamodels and filter on adult/senior evidence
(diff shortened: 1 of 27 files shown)


kzuberihmf (commented):
Not sure where the new age groups table get's deleted, should somehow follow from EvidenceDAO.deleteAll()?

kzuberihmf (approved):
Looks good to me otherwise. BTW I looked over this once, didn't notice any issues, then asked an LLM to review and it was pointing out the details I mentioned 😄. Probably in the future we will adopt such review as part of our regular practice, I do find it helpful!

kduyvesteyn:
> Looks good to me otherwise. BTW I looked over this once, didn't notice any issues, then asked an LLM to review and it was pointing out the details I mentioned 😄. Probably in the future we will adopt such review as part of our regular practice, I do find it helpful!

Nice!

PR #73, opened 2026-06-15
OPR-332: Support manually curated trials

kduyvesteyn:
The initial setup was by exposing the full SERVE datamodel but I didn't like that setup (it's impossible to manually configure) so I have simplified the input datamodel. 

The key decisions I've made:
 - There are no COMBINED events. If a trial is valid for multiple indications and/or molecular criteria we expect multiple lines (which are then aggregated). 
 - The `actionableGene` and `actionableEvent` have to adhere to CKB formatting rules and are then extracted as such. I imagine this offers quite a natural interface for e.g. medical team to configure molecular rules. 

PR has been reviewed by CLAUDE and have added a small README about key assumptions made when ingesting this new knowledgebase.

Has been tested using nextflow pipeline (and one fake curated trial)

### DIFF
--- a/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
+++ b/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
@@ -27,6 +27,9 @@ public abstract class HartwigTrialEntry {
     public abstract String cancerType();
 
     @NotNull
+    public abstract String cancerTypeDoid();
+
+    @NotNull
     public abstract String molecularCriteriumType();
 
     @NotNull
--- a/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
+++ b/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
@@ -32,7 +32,10 @@ public abstract class HartwigTrialEntry {
     public abstract String cancerTypeDoid();
 
     @NotNull
-    public abstract String molecularCriterium();
+    public abstract String actionableGene();
+    
+    @NotNull
+    public abstract String actionableEvent();
     
     @NotNull
     public abstract String url();
--- a/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
+++ b/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
@@ -1,5 +1,7 @@
 package com.hartwig.serve.sources.hartwig.trial;
 
+import java.time.LocalDate;
+
 import com.hartwig.serve.datamodel.trial.GenderCriterium;
 
 import org.immutables.value.Value;
@@ -10,6 +12,9 @@ import org.jetbrains.annotations.Nullable;
 @Value.Style(passAnnotations = { NotNull.class, Nullable.class })
 public abstract class HartwigTrialEntry {
 
+    @NotNull
+    public abstract LocalDate date();
+    
     @NotNull
     public abstract String nctId();
 
--- a/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
+++ b/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
@@ -1,5 +1,7 @@
 package com.hartwig.serve.sources.hartwig.trial;
 
+import com.hartwig.serve.datamodel.trial.GenderCriterium;
+
 import org.immutables.value.Value;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
@@ -14,23 +16,20 @@ public abstract class HartwigTrialEntry {
     @NotNull
     public abstract String title();
 
-    @NotNull
+    @Nullable
     public abstract String acronym();
     
     @NotNull
     public abstract String country();
     
     @Nullable
-    public abstract String genderCriterium();
+    public abstract GenderCriterium genderCriterium();
 
     @NotNull
     public abstract String cancerType();
 
     @NotNull
     public abstract String cancerTypeDoid();
-
-    @NotNull
-    public abstract String molecularCriteriumType();
 
     @NotNull
     public abstract String molecularCriterium();
--- a/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
+++ b/algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
@@ -1,13 +1,5 @@
 package com.hartwig.serve.sources.hartwig.trial;
 
-import java.util.List;
-import java.util.Set;
-
-import com.hartwig.serve.datamodel.common.Indication;
-import com.hartwig.serve.datamodel.molecular.MolecularCriterium;
-import com.hartwig.serve.datamodel.trial.Country;
-import com.hartwig.serve.datamodel.trial.GenderCriterium;
-
 import org.immutables.value.Value;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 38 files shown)


algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
```diff
@@ -0,0 +1,47 @@
...
+import org.jetbrains.annotations.Nullable;
+
+@Value.Immutable
+@Value.Style(passAnnotations = { NotNull.class, Nullable.class })
+public abstract class HartwigTrialEntry {
+
+    @NotNull
+    public abstract LocalDate date();
+    
+    @NotNull
```
LiekeSchoenmaker, line 18: Should this not nullable, because not all ACTIN trial has a NCT ID?

algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
```diff
@@ -0,0 +1,47 @@
...
+
+import com.hartwig.serve.datamodel.trial.GenderCriterium;
+
+import org.immutables.value.Value;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+@Value.Immutable
+@Value.Style(passAnnotations = { NotNull.class, Nullable.class })
+public abstract class HartwigTrialEntry {
```
LiekeSchoenmaker, line 13: Is there none therapy available per study?

algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialExtractor.java
```diff
@@ -0,0 +1,141 @@
...
+
+    HartwigTrialExtractor(@NotNull MolecularCriteriumExtractor molecularCriteriumExtractor) {
+        this.molecularCriteriumExtractor = molecularCriteriumExtractor;
+    }
+
+    @NotNull
+    public ExtractionResult extract(@NotNull List<HartwigTrialEntry> entries) {
+        Set<ActionableTrial> actionableTrials = new HashSet<>();
+        ProgressTracker tracker = new ProgressTracker("Hartwig Trials", entries.size());
+
```
LiekeSchoenmaker, line 46: Does this work when a trial hasn't a NCT ID?

LiekeSchoenmaker (commented):
Looks good to me. 
However, If I reminder correctly, not all ACTIN studies has a NCT ID. In this datamodel the NCT ID is Notnull. Is this as expected?

kduyvesteyn:
> Looks good to me. However, If I reminder correctly, not all ACTIN studies has a NCT ID. In this datamodel the NCT ID is Notnull. Is this as expected?

Good point! I picked @NotNull since SERVE expects all trials to have an NCT ID. Not all ACTIN trials have an NCT ID indeed so this doesn't match. The 2 options are:
 - Make NCT ID nullable in SERVE -> feels like quite a big change with uncertain impact.
 - Allow a fake ID to be filled in the manual trial entries -> depends on the nctId not being used in actual logic. 

Do you have a preference @LiekeSchoenmaker ?

algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
kduyvesteyn: I kept therapies out of the manual trial config assuming it is not important. If you think it should be added I will add it? Didn't look up its usage in PROTECT.

algo/src/test/resources/hartwig/hartwig_curated_trials.tsv
```diff
@@ -0,0 +1,2 @@
+date	nctId	title	acronym	country	genderCriterium	cancerType	cancerTypeDoid	actionableGene	actionableEvent	url
+20-Jan-2026	NCT00001	This is test trial 1	TRIAL-1	Netherlands	BOTH	Colorectal Cancer	123	-	EML4 - ALK Fusion	https://url.com
```
kzuberihmf: Minor, is `ELM4 - ALK Fusion` correct formatting? I see unit tests have no space around the genes, like `ELM4-ALK fusion`.

algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialExtractor.java
```diff
@@ -0,0 +1,141 @@
...
+        return extractSingleValue(entriesForNctId, HartwigTrialEntry::acronym);
+    }
+
+    @NotNull
+    private static Country extractCountry(@NotNull List<HartwigTrialEntry> entriesForNctId) {
+        return extractSingleValueNotNull(entriesForNctId, entry -> ImmutableCountry.builder().name(entry.country()).build());
+    }
+
+    @NotNull
+    private static GenderCriterium extractGenderCriterium(@NotNull List<HartwigTrialEntry> entriesForNctId) {
```
kzuberihmf: GenderCriterium is nullable in HartwigTrialEntry.

kzuberihmf (commented):
Seems reasonable to me.

algo/src/main/java/com/hartwig/serve/sources/hartwig/trial/HartwigTrialEntry.java
LiekeSchoenmaker: therapy names/drug names should be used in protect because this value will be displayed in the OncoAct report

LiekeSchoenmaker:
> > Looks good to me. However, If I reminder correctly, not all ACTIN studies has a NCT ID. In this datamodel the NCT ID is Notnull. Is this as expected?
> 
> Good point! I picked @NotNull since SERVE expects all trials to have an NCT ID. Not all ACTIN trials have an NCT ID indeed so this doesn't match. The 2 options are:
> 
> * Make NCT ID nullable in SERVE -> feels like quite a big change with uncertain impact.
> * Allow a fake ID to be filled in the manual trial entries -> depends on the nctId not being used in actual logic.
> 
> Do you have a preference @LiekeSchoenmaker ?

Maybe a fake ID is the best way in this case. One idea for the idea is using e.g. Hartwig1, Hartwig2. But maybe you have another preference?

kduyvesteyn:
> > > Looks good to me. However, If I reminder correctly, not all ACTIN studies has a NCT ID. In this datamodel the NCT ID is Notnull. Is this as expected?
> > 
> > 
> > Good point! I picked @NotNull since SERVE expects all trials to have an NCT ID. Not all ACTIN trials have an NCT ID indeed so this doesn't match. The 2 options are:
> > 
> > * Make NCT ID nullable in SERVE -> feels like quite a big change with uncertain impact.
> > * Allow a fake ID to be filled in the manual trial entries -> depends on the nctId not being used in actual logic.
> > 
> > Do you have a preference @LiekeSchoenmaker ?
> 
> Maybe a fake ID is the best way in this case. One idea for the idea is using e.g. Hartwig1, Hartwig2. But maybe you have another preference?

The current patient reporter creates a URL out of the NCT ID indeed, but still think that faking it is the best way forward. I expect it to be rare that we need to configure a manual trial anyway, and if we configure something manually it should still have an NCT ID in most cases. 

And even in worst case scenario (faking it), the reporter would not break - it would only generate a link that doesn't exist. We could potentially reduce some assumptions in the patient reporter for this but even that I think is not worth it (my expectation is that soon after this release we will switch to ACTIN source for trials rather than SERVE)

LiekeSchoenmaker:
> > > > Looks good to me. However, If I reminder correctly, not all ACTIN studies has a NCT ID. In this datamodel the NCT ID is Notnull. Is this as expected?
> > > 
> > > 
> > > Good point! I picked @NotNull since SERVE expects all trials to have an NCT ID. Not all ACTIN trials have an NCT ID indeed so this doesn't match. The 2 options are:
> > > 
> > > * Make NCT ID nullable in SERVE -> feels like quite a big change with uncertain impact.
> > > * Allow a fake ID to be filled in the manual trial entries -> depends on the nctId not being used in actual logic.
> > > 
> > > Do you have a preference @LiekeSchoenmaker ?
> > 
> > 
> > Maybe a fake ID is the best way in this case. One idea for the idea is using e.g. Hartwig1, Hartwig2. But maybe you have another preference?
> 
> The current patient reporter creates a URL out of the NCT ID indeed, but still think that faking it is the best way forward. I expect it to be rare that we need to configure a manual trial anyway, and if we configure something manually it should still have an NCT ID in most cases.
> 
> And even in worst case scenario (faking it), the reporter would not break - it would only generate a link that doesn't exist. We could potentially reduce some assumptions in the patient reporter for this but even that I think is not worth it (my expectation is that soon after this release we will switch to ACTIN source for trials rather than SERVE)

we can maybe generate only a url when there is a NCT available...otherwise no link is provided but only the fake ID on the report

kzuberihmf (approved):

LiekeSchoenmaker (approved):
Looks good to me

PR #74, opened 2026-06-15
OPR-450: Bump java and maven versions

kduyvesteyn:
(Also bumped to the correct year in our LICENSE)

@kzuberihmf Would you know if java21 would work for SERVE given the docker we create to run transvar and which requires python2 (see `Dockerfile` in algo)? I will just give it a try but if you already know the answer or could provide guidance that would be appreciated!

UPDATE: All code looks good, cloud build works and currently running full regression run ([kd-20260615-3](https://console.cloud.google.com/batch/jobs?project=actin-shared))

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -18,10 +18,8 @@
 
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
-
-        <java.version>21</java.version>
-        <maven.compiler.source>${java.version}</maven.compiler.source>
-        <maven.compiler.target>${java.version}</maven.compiler.target>
+        
+        <maven.compiler.release>21</maven.compiler.release>
 
         <maven.compiler.plugin.version>3.13.0</maven.compiler.plugin.version>
         <maven.assembly.plugin.version>3.6.0</maven.assembly.plugin.version>
--- a/pom.xml
+++ b/pom.xml
@@ -19,14 +19,17 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <maven.compiler.source>21</maven.compiler.source>
-        <maven.compiler.target>21</maven.compiler.target>
+        <java.version>21</java.version>
+        <maven.compiler.source>${java.version}</maven.compiler.source>
+        <maven.compiler.target>${java.version}</maven.compiler.target>
 
-        <maven.compiler.plugin.version>3.8.1</maven.compiler.plugin.version>
-        <maven.assembly.plugin.version>3.0.0</maven.assembly.plugin.version>
-        <maven.jar.plugin.version>2.4</maven.jar.plugin.version>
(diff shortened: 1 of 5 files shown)


kzuberihmf:
> (Also bumped to the correct year in our LICENSE)
> 
> @kzuberihmf Would you know if java21 would work for SERVE given the docker we create to run transvar and which requires python2 (see `Dockerfile` in algo)? I will just give it a try but if you already know the answer or could provide guidance that would be appreciated!

I think this is unlikely to work, the eclipse-temurin images stopped supporting python2 at some point. We did manage to update actin-molecular to newer java, and it also depends on transvar, so suggest you try copying the strategy there: https://github.com/hartwigmedical/actin/blob/master/molecular/transvar.Dockerfile.

kduyvesteyn:
> https://github.com/hartwigmedical/actin/blob/master/molecular/transvar.Dockerfile

Found out that the jammy version of temurin is still based on ubuntu 22.04, so that works. I did have to pull in pip via a curl, is that "clean"? 

In general should there be a difference between these 2 images (both j21 + python2)? @kzuberihmf Which version would be you prefer (I'll just duplicate that version then)

pom.xml
```diff
@@ -19,14 +19,17 @@
     <properties>
         <project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 
-        <maven.compiler.source>11</maven.compiler.source>
-        <maven.compiler.target>11</maven.compiler.target>
+        <java.version>21</java.version>
+        <maven.compiler.source>${java.version}</maven.compiler.source>
+        <maven.compiler.target>${java.version}</maven.compiler.target>
```
ArnoHMF: It's recommend to use <maven.compiler.release>${java.version}</maven.compiler.release> instead from Java 9+

ArnoHMF (approved):
LGTM

kzuberihmf:
Nice find with this jammy image. You should be able to avoid using curl to fetch an external pip by just including `python-pip` in the apt-get install line, which is pip2 for this distribution.

The rest of actin using temurin images, so I should probably switch to this solution for actin-molecular for consistency. 

When you test the image, perhaps watch for any strange `locale` related issues, that setting is something we added to the molecular version, though I don't recall the details of the issue or what triggers it.

kzuberihmf (approved):

kduyvesteyn:
> Nice find with this jammy image. You should be able to avoid using curl to fetch an external pip by just including `python-pip` in the apt-get install line, which is pip2 for this distribution.
> 
> The rest of actin using temurin images, so I should probably switch to this solution for actin-molecular for consistency.
> 
> When you test the image, perhaps watch for any strange `locale` related issues, that setting is something we added to the molecular version, though I don't recall the details of the issue or what triggers it.

Ah good point re `python-pip`, will give that a try.

The regression run has finished by now and looks all good, can't find anything specific to locale in the logs.

PR #75, opened 2026-06-25
ACTIN-3001: Remove config for full vs partial exonic delete

kduyvesteyn:
I've reviewed the impact of this change and I think this is all there is to it. The CKB events now get classified correctly (see updates to unit test for `CkbEventTypeExtractorTest`), and the only CKB event that is still configured as both an exonic delete as well as exonic fusion is `KIT exon 11`. 

With the removal of the config in this ticket this will always be assumed to be a full exonic delete, we simply don't support intra-exon fusions anymore (which are unlikely to be called anyways).

### DIFF
--- a/algo/src/main/java/com/hartwig/serve/extraction/fusion/FusionAnnotationConfig.java
+++ b/algo/src/main/java/com/hartwig/serve/extraction/fusion/FusionAnnotationConfig.java
@@ -11,8 +11,6 @@ final class FusionAnnotationConfig {
 
     public static final Map<String, FusionPair> EXONIC_FUSIONS_MAP = exonicFusionMap();
 
-    public static final Map<String, ExonicDelDupType> DEL_DUP_TYPE_PER_GENE = exonicDelDupTypes();
-
     @NotNull
     private static Map<String, FusionPair> exonicFusionMap() {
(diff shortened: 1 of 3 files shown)


cbruel (commented):
Totally forgot about this one, thanks for picking it up!

If I only add the two new tests to `CkbEventTypeExtractorTest`, and don't make any other changes to the code, your new tests also pass? That seems unexpected to me (i.e. either testing is insufficient or your changes don't have an impact?) but maybe miss Bruel is struck by the heat

kduyvesteyn:
> Totally forgot about this one, thanks for picking it up!
> 
> If I only add the two new tests to `CkbEventTypeExtractorTest`, and don't make any other changes to the code, your new tests also pass? That seems unexpected to me (i.e. either testing is insufficient or your changes don't have an impact?) but maybe miss Bruel is struck by the heat

You are not struck by the heat, I got confused as well when I looked into this! The config that we deleted _only impacts how fusion pairs that are exonic del-dups are interpreted_, so the classification by itself isn't changed. I started with adding those tests indeed assuming they would fail but I misunderstood the code.

Side note - this feature wasn't tested very well apparently to begin with since it shouldn't be possible to remove this type of code without having to change a unit test :)

cbruel (approved):

PR #76, opened 2026-07-08
OPR-468: Force deterministic order of events in SERVE output

kduyvesteyn:
When checking the SERVE output of recent runs I discovered that there is some randomness in the sorting. Turned out that was caused by various issues on various levels. I think I got them all but still testing (one run per day). 

Since I am leaving for holiday soon, already opening this PR. Even if I didn't catch all causes for randomness I feel this PR is an improvement anyhow. 

(Side note: I considered making the datamodel enforce sorting but decided not to since there is no logical ordering anyways, and clients of the datamodel should not assume there is an ordering - the sorting is therefore an implementation detail simply to help with regression testing)

EDIT: I did 2 runs on 8th of July with same input and it produced identical output, which I hadn't succeeded in before. Makes me more confident I got all the randomness ironed out!

### DIFF
--- a/algo/src/main/java/com/hartwig/serve/extraction/ExtractionFunctions.java
+++ b/algo/src/main/java/com/hartwig/serve/extraction/ExtractionFunctions.java
@@ -55,17 +55,21 @@ public final class ExtractionFunctions {
                 }
                 unconsolidatedKnownEventsBuilder.from(result.knownEvents());
             }
-            if (result.evidences() != null) {
+            
+            List<EfficacyEvidence> evidences = result.evidences();
+            if (evidences != null) {
                 if (unconsolidatedEvidences == null) {
                     unconsolidatedEvidences = Lists.newArrayList();
                 }
-                unconsolidatedEvidences.addAll(result.evidences());
+                unconsolidatedEvidences.addAll(evidences);
             }
-            if (result.trials() != null) {
+            
+            List<ActionableTrial> trials =  result.trials();
+            if (trials!= null) {
                 if (unconsolidatedTrials == null) {
                     unconsolidatedTrials = Lists.newArrayList();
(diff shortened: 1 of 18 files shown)


kzuberihmf (approved):
Couple of notes:

* A quick agent review suggests there are still places that pass sets that will be compared by iteration order (as your comment in compareSetOfComparable() says, it doesn't defensively sort itself). But I didn't attempt to verify, and since this seems like an improvement I think its still good to apply
* There was a previous attempt to enforce sorting at the json serialization step, see in ServeJson.java. Perhaps some of this is redundant now. But it makes me wonder why it wasn't sufficient before, and it may be that same behaviour as for compareSetOfComparable().  That is, nested sets being compared and iteration order ending up being decisive.

kduyvesteyn:
At least our agents are consistent since mine suggested to "fix" the Set comparing as well which I initially took over but then realized it is not the right solution. Our datamodel doesn't enforce a set to be sorted, so why should we assume and even enforce sorting inside the comparison. In the end, if the datamodel doesn't enforce sorting then {A, B} is really different from {B, A}!

I added a comment on the ServeJson sorting mechanism but felt fine to leave in (no super strong opinion btw). The reason why it wasn't sufficient (I think) is because of the recursiveness of this problem. If we sort a list of efficacy evidences that will go fine but if their underlying molecular criteria are not sorted first then we sort on the wrong thing. Or put differently: we have to make sure we sort objects that have each of their properties sorted first, etc. 

In my experiments I mainly noticed things went non-deterministic with combined molecular profiles, suggesting that the Set of `MolecularCriterium` wasn't sorted for all evidences / trials, so when you don't sort a list of evidences you get non-deterministic behaviour. Hope this makes sense!

